PR #39, opened 2026-07-04
Fix #1: enforce auth + role checks on all API routes

TusharW4ni:
## What was broken
25 of 33 API endpoints had **zero authentication** — any unauthenticated request could read, modify, or delete rides, clients, volunteers, users, and admins (#1). That included `POST /api/post/admins`, so anyone could promote themselves to ADMIN (#2).

## What changed
- **`server/middleware/auth.ts`** (new) — global Nitro middleware gating every `/api` route except better-auth's own `/api/auth/**`:
  - No session → **401**
  - `ADMIN` → full access
  - `VOLUNTEER` → reads + own self-service writes (ride signup/unsignup, `put/volunteers/bySession/*`); any other write → **403**
  - `CLIENT` / unmapped role → **403** (the internal API is admin/volunteer facing)
- **`server/utils/auth.ts`** — the login before-hook now queries Prisma directly instead of `$fetch`ing `/api/get/users/byEmail`. That loopback call would otherwise hit the new middleware pre-session, always error, and silently disable the "contact admin to be added" login gate. This also resolves #20.
- Per-handler checks (e.g. `signup.ts` verifying ride ownership) are unchanged — the middleware is the coarse gate in front of them.

## Verification
- `tests/e2e/auth-middleware.test.ts` (new, 7 cases): unauthenticated reads/writes → 401; volunteer → 403 on `post/admins` and `delete/volunteers`; client → 403 on `get/rides`; volunteer and admin happy paths; unknown-email login gate still returns 400.
- **Revert-check**: with the middleware removed, the 4 core enforcement tests fail; restored, they pass — the tests genuinely pin the hole.
- Full suite: 10/10 green. `pnpm build` succeeds.
- Rate limiting is disabled **only** in the e2e env (`DISABLE_RATE_LIMIT=true` in `vitest.config.ts`); production builds keep better-auth's default rate limiting on.

## Notes for reviewer
- This is a `P0: critical` security change to the production auth surface — merging deploys. Please eyeball the role model before merge.
- #3 (PII overfetching — volunteers can still *read* all rides/clients) is intentionally **not** addressed here; it needs per-user query scoping and has its own issue.

Fixes #1
Fixes #2
Fixes #20

PR #40, opened 2026-07-04
Fix #3: scope API data to the requesting user

### LINKED ISSUE #3: [HIGH] Client-Side Overfetching of PII Data
### Location
`app/pages/rides/index.vue` and `app/pages/people.vue`

### Description
When a `VOLUNTEER` views their dashboard, the application makes a blanket request to `/api/get/rides`, `/api/get/volunteers`, and `/api/get/clients`. It downloads the **entire database** of people and rides (including full names, phone numbers, and addresses) to the client's browser, and uses frontend JavaScript to hide records that aren't assigned to that specific volunteer. This is a massive PII data leak.

### Recommended Fix
Move filtering to the server backend. Pass user authentication contexts into API queries so the endpoint only returns data that the requestor is authorized to view.

TusharW4ni:
## What was broken

When a VOLUNTEER viewed their dashboard, the app fetched `/api/get/rides`, `/api/get/clients`, and `/api/get/volunteers` in full — downloading the **entire** database of rides and people (names, phone numbers, home addresses) to the browser — then hid non-relevant records with client-side JavaScript. That is a PII leak: any authenticated volunteer could read every client's address and every other volunteer's contact info straight from the network responses. The global auth middleware (PR #39) gated *access* to these routes but did not *scope* what they return.

## What changed

Server-side data scoping (the layer after authentication):

- **`server/api/get/rides/index.ts`** — a non-ADMIN now receives only rides that are available to sign up for (`status = CREATED`) **or** assigned to them (`volunteer.userId === session.user.id`). ADMINs still receive every ride. This keeps the volunteer rides dashboard working (they still see available + their own rides) while no longer leaking other volunteers' assignments and the attached client PII.
- **`server/api/get/clients/index.ts`** and **`server/api/get/volunteers/index.ts`** — now **admin-only** (`403` otherwise). These bulk PII rosters are only consumed by admin-facing UI (the create/edit-ride volunteer picker and the people-management page).
- **`server/middleware/auth.ts`** — stashes the already-resolved session in `event.context.auth` so handlers can scope without a second session lookup.
- **`server/utils/authz.ts`** (new) — small auto-imported `requireAdmin(event)` / `getAuth(event)` helpers.

No auth config, schema, CI, docker, or dependency changes.

## Verification

New regression test `tests/e2e/pii-scoping.test.ts`:
- `VOLUNTEER only receives available (CREATED) rides plus their own` — asserts every returned ride is CREATED or assigned to the requester, and that no foreign volunteer's ride appears.
- `VOLUNTEER cannot bulk-download the client roster (PII)` → expects `403`.
- `VOLUNTEER cannot bulk-download the volunteer roster (PII)` → expects `403`.
- `ADMIN still receives every ride (full dataset)` and `ADMIN can still read the client and volunteer rosters` — confirms admin behavior is unchanged.

Confirmed the three volunteer-scoping tests **fail on the unmodified code** for the right reasons (a foreign COMPLETED ride leaked to Bob; `/api/get/clients` and `/api/get/volunteers` returned full rosters) and **pass** after the fix. Full suite green (15 tests), `pnpm build` succeeds.

## Reviewer note

`app/pages/rides/index.vue` and `app/pages/rides/[id].vue` still call `useFetch('/api/get/volunteers')` unconditionally, and `rides/index.vue` also calls `/api/get/clients`. For a volunteer these now return `403`; the data was only ever used inside `v-if="isAdmin"` blocks (the volunteer picker / create-ride modal), so `useFetch` just surfaces an error ref and the computed option lists fall back to `[]` — the pages render fine and volunteer-facing functionality is unaffected. A follow-up could gate those `useFetch` calls behind `isAdmin` to avoid the benign 403s, but that is frontend cleanup out of scope here.

Fixes #3

### DIFF
--- a/server/api/get/clients/index.ts
+++ b/server/api/get/clients/index.ts
@@ -1,6 +1,10 @@
 import { prisma } from '../../../utils/prisma'
 
-export default defineEventHandler(async () => {
+export default defineEventHandler(async (event) => {
+  // Bulk client roster carries full PII (name/phone/address) and is only used
+  // by admin-facing UI — keep it admin-only (issue #3).
+  requireAdmin(event)
+
   return await prisma.client.findMany({
     include: {
       user: true,
--- a/server/api/get/rides/index.ts
+++ b/server/api/get/rides/index.ts
@@ -1,10 +1,28 @@
+import { Prisma } from '../../../../prisma/generated/client'
 import { prisma } from '../../../utils/prisma'
 
 export default defineEventHandler(async (event) => {
   const query = getQuery(event)
   const sort = query.sort === 'desc' ? 'desc' : 'asc'
 
+  // The global auth middleware guarantees a session and stashes it here.
+  const session = event.context.auth as
+    | { user: { id: string; role?: string } }
+    | undefined
+  const role = session?.user?.role
+
+  // Data scoping (issue #3): a VOLUNTEER must only receive rides that are
+  // available to sign up for (CREATED) or that are assigned to them — never
+  // other volunteers' rides or the full roster of client PII. ADMINs see all.
+  let where: Prisma.RideWhereInput | undefined
+  if (role !== 'ADMIN') {
+    where = {
+      OR: [{ status: 'CREATED' }, { volunteer: { userId: session?.user?.id } }],
+    }
+  }
+
   return await prisma.ride.findMany({
+    where,
     include: {
       client: {
         include: {
@@ -22,4 +40,3 @@ export default defineEventHandler(async (event) => {
     },
   })
 })
-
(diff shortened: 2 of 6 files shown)


TusharW4ni:

| Gate | Result |
|---|---|
| Independent reviewer full suite | ✅ 15/15 green (re-run from scratch in a clean worktree) |
| Revert-check | ✅ 3 volunteer-scoping tests fail without the fix, pass with it |
| Scope | ✅ 3 endpoints + middleware + authz helper + tests; net +49/-2 server |
| Risk files | ✅ none (no auth-config/schema/CI/docker/deps) |
| `/review` findings | ✅ one PLAUSIBLE fail-open in the rides `where` clause — **fixed** in 8209a5d (fail closed when userId absent), suite re-verified green |
| Priority | P1 (not P0) → eligible for auto-merge |

Follow-up filed separately: `get/rides/byId/[id]` and other single-record endpoints are not yet user-scoped (same vuln class, outside #3's list-endpoint scope).

PR #43, opened 2026-07-04
Fix #41: scope single-record & admin-only endpoints

### LINKED ISSUE #41: Single-record & metrics endpoints still leak PII to non-admins
### Discovered during review of #3 (PR #40)

#3 scoped the **list** endpoints (`get/rides`, `get/clients`, `get/volunteers`). But other authenticated endpoints still return full data to any logged-in non-admin volunteer:

- `server/api/get/rides/byId/[id].ts` — a volunteer can fetch **any** ride by id, including the client's name/phone/address, even rides not theirs and not available.
- `server/api/get/users/*`, `get/admins`, `get/addresses`, `get/metrics/*` — no per-user scoping.

### Recommended fix
Apply the same scoping as #3: for `byId`, a non-admin should only retrieve a ride that is CREATED or assigned to them (reuse the `where` logic / `getAuth` helper from `server/utils/authz.ts`). Audit the remaining endpoints and lock down or scope each.

Same vulnerability class as #3; split out to keep that PR focused.

TusharW4ni:
## What was broken

#3 scoped the **list** endpoints (`get/rides`, `get/clients`, `get/volunteers`), but single-record and other authenticated endpoints still returned full data to any logged-in non-admin VOLUNTEER. A volunteer could fetch **any** ride by id — including client name/phone/address — even for rides not theirs and not available; and `get/users/*`, `get/admins`, `get/addresses`, `get/metrics/*` had no per-user scoping (e.g. a volunteer could pull the completion-rate/hours metrics and the top-riders list of client names).

## What changed

Reuses the existing `getAuth`/`requireAdmin` helpers in `server/utils/authz.ts` (from #3) — no new mechanism.

**Record-scoped** (mirrors `get/rides/index.ts`):
- `server/api/get/rides/byId/[id].ts` — a non-admin may retrieve a ride **only** if it is `CREATED` (available to sign up) or assigned to them; otherwise **404**. 404 rather than 403 is deliberate so the endpoint does not reveal that a ride the caller isn't authorized to see exists. ADMIN sees any ride; the not-found case for ADMIN still returns `null` as before.

**Made admin-only** (`requireAdmin` → 403 for non-admins). Consumers checked:
- `server/api/get/users/index.ts`, `byId/[id].get.ts`, `byEmail/[email].get.ts` — no frontend consumers; pure admin management data.
- `server/api/get/admins/index.ts` — consumer `app/pages/people.vue`, an admin-only page (route guard redirects volunteers).
- `server/api/get/addresses/index.ts` — consumer is the create-ride autocomplete on `app/pages/rides/index.vue`, which is admin-gated (`v-if="isAdmin"`); the sibling `clients`/`volunteers` fetches on that same page were already made admin-only in #3, so the shared page already tolerates a 403 on an unconditional `useFetch`.
- `server/api/get/metrics/completionRate|hours|topRiders/index.ts` — consumer `app/pages/index.vue`, the admin dashboard guarded by the route middleware from #4.

## Verification

Regression test: `tests/e2e/record-scoping.test.ts`.

**Revert-check (test proves the bug):** with the source fix stashed (test only), the 9 security assertions fail for the right reason — the volunteer receives the protected data instead of a rejection, e.g.:
- `byId` foreign ride: `promise resolved "{ …(16) }" instead of rejecting` (expected 404)
- `/api/get/metrics/completionRate`: `promise resolved "{ percentage: 33, total: 9, …(1) }" instead of rejecting` (expected 403)
- `/api/get/metrics/topRiders`: `promise resolved "[ { name: 'Sarah Connor', …(1) }, …(2) ]"` (expected 403)
- `/api/get/users`, `/api/get/admins`, `/api/get/addresses`, `get/users/byId`, `get/users/byEmail` all resolved with data instead of 403.

With the fix restored, all pass. **Full suite: 39 passed (5 files). `pnpm build` succeeds.**

Test coverage added:
- VOLUNTEER gets 404 fetching another volunteer's non-available ride by id; CAN fetch an available (`CREATED`) ride and their own assigned ride; ADMIN can fetch any.
- VOLUNTEER gets 403 on each admin-only endpoint; ADMIN still succeeds.

Fixes #41

### DIFF
--- a/server/api/get/addresses/index.ts
+++ b/server/api/get/addresses/index.ts
@@ -1,6 +1,11 @@
 import { prisma } from '../../../utils/prisma'
 
 export default defineEventHandler(async (event) => {
+  // Client addresses are PII (issue #41). Only the create-ride autocomplete
+  // uses this, and that form is admin-only (rides/index.vue gates it behind
+  // isAdmin; #3 already made the sibling clients/volunteers lists admin-only).
+  requireAdmin(event)
+
   const query = getQuery(event)
   const search = query.search as string
 
--- a/server/api/get/admins/index.ts
+++ b/server/api/get/admins/index.ts
@@ -1,6 +1,10 @@
 import { prisma } from '../../../utils/prisma'
 
-export default defineEventHandler(async () => {
+export default defineEventHandler(async (event) => {
+  // Admin roster is management data / PII (issue #41). Consumer: people.vue,
+  // an admin-only page (volunteers are redirected by the route guard).
+  requireAdmin(event)
+
   return await prisma.user.findMany({
     where: {
       role: 'ADMIN',
--- a/server/api/get/metrics/completionRate/index.ts
+++ b/server/api/get/metrics/completionRate/index.ts
@@ -1,6 +1,10 @@
 import { prisma } from '../../../../utils/prisma'
 
 export default defineEventHandler(async (event) => {
+  // Dashboard metrics are admin-only (issue #41). Consumer: index.vue, the
+  // admin dashboard (volunteers are redirected by the route guard, #4).
+  requireAdmin(event)
+
   const query = getQuery(event)
   const startDate = query.startDate ? new Date(String(query.startDate)) : undefined
   const endDate = query.endDate ? new Date(String(query.endDate)) : undefined
--- a/server/api/get/metrics/hours/index.ts
+++ b/server/api/get/metrics/hours/index.ts
@@ -1,6 +1,10 @@
 import { prisma } from '../../../../utils/prisma'
 
 export default defineEventHandler(async (event) => {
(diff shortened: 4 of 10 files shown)


TusharW4ni:

| Gate | Result |
|---|---|
| Independent reviewer full suite | ✅ 39/39 green (clean-worktree re-run) |
| Revert-check | ✅ 9 distinct assertions fail without the fix, all pass with it |
| Scope | ✅ 9 handlers + 1 test, net +46 |
| Risk files | ✅ none |
| Critical login-regression check | ✅ before-hook queries Prisma directly (post #39/#20); no pre-session caller of the now-admin-only `byEmail`/`users`/`metrics` endpoints |
| `/review` findings | ✅ `byId` throws 404 before returning any body; rule matches #3; fails closed on missing userId |
| Priority | P1 → eligible for auto-merge |

PR #44, opened 2026-07-06
Fix #31: validate request bodies on write endpoints

### LINKED ISSUE #31: No Server-Side Input Validation / Mass Assignment on Write Endpoints
### Location
All write endpoints; worst case is `server/api/put/rides/[id].ts`

### Description
No server endpoint validates its request body against a schema (zod is a dependency but is only used in frontend forms). The sharpest instance is `put/rides/[id].ts`:

```ts
const updateData: any = { ...body }   // entire client-controlled body
await prisma.ride.update({ where: { id }, data: updateData })
```

Consequences:
- **Mass assignment** — a caller can set any `Ride` column: overwrite `clientId`, `createdAt`, `totalRideTime`, forge `pickupDisplay`, etc.
- **500s on unknown keys** — any property Prisma doesn't recognize throws `PrismaClientValidationError`, so a slightly-wrong client payload crashes the request with an opaque error.
- Type coercion issues elsewhere (see #15 empty-string phones, #10 missing `totalRideTime`) are all symptoms of the same missing validation layer.

### Recommended Fix
Define zod schemas per endpoint in `server/` (they can be shared with the frontend since zod is already isomorphic here), parse with `schema.safeParse(await readBody(event))`, return 400 with field errors on failure, and build the Prisma `data` object only from parsed, whitelisted fields. Start with the ride/client/volunteer PUT/POST routes.

Complements #1 (authn) — that controls *who* can call the API; this controls *what* they can send.

TusharW4ni:
## What was broken

`server/api/put/rides/[id].ts` spread the entire client-controlled request body straight into `prisma.ride.update` (`const updateData: any = { ...body }`). This allowed **mass assignment** of any `Ride` column (overwrite `clientId`, forge `createdAt`/`totalRideTime`/`pickupDisplay`, etc.) and, because Prisma rejects unrecognized keys, any slightly-wrong payload crashed the request with an opaque **500** (`PrismaClientValidationError`).

## What changed

- New shared helper `server/utils/validation.ts` — `readValidatedBody(event, schema)` reads the body, `safeParse`s it, and throws a **400** with per-field error messages on failure (Nitro auto-imports `server/utils/*`).
- `server/api/put/rides/[id].ts` now defines a `.strict()` zod schema of only the updatable fields (`status` enum CREATED|ASSIGNED|COMPLETED, `volunteerId` nullable, `scheduledTime`, `pickupTime` nullable, `totalRideTime`, `notes`, `pickup/dropoffDisplay`), parses the body through the helper, and builds the Prisma `data` object **only** from parsed fields. Unknown/forbidden keys now 400.

**Preserved behavior:** `volunteerId` empty-string → `null` unassign, `scheduledTime`/`pickupTime` Date conversion (null/empty clears `pickupTime`), and the RIDE_COMPLETED / (dead) RIDE_CANCELLED notification side effects — all now read from the validated object.

## Scope

- **Covered:** `PUT /api/put/rides/[id]` (the sharpest instance — the only true mass-assignment sink; `post/rides/index.ts` already whitelists its `data`).
- **Deferred (follow-up):** applying `readValidatedBody` + schemas to `post/rides/index.ts`, `put/clients/[id]`, `put/volunteers/[id]` and other writes for type-level input validation. These do not mass-assign today (they build `data` explicitly), so kept out to preserve reviewability. The new helper makes those a small drop-in.

No changes to the RideStatus enum, prisma schema, auth, CI, docker, or dependencies (zod was already present).

## Verification

New regression test `tests/e2e/ride-input-validation.test.ts`:
- `rejects a forbidden field (mass assignment) with 400 and does not persist it`
- `rejects an unknown/bogus key with 400 instead of an opaque 500`
- `accepts a valid update (status + notes) and returns 200`
- `accepts unassigning a volunteer via empty-string volunteerId (preserved behavior)`

**Before the fix** the two mass-assignment/unknown-key cases failed because the endpoint returned `500 Server Error` instead of `400` (e.g. `expected FetchError: [PUT] ... 500 Server Error to match object { statusCode: 400 }`). **After the fix** all 4 pass.

- `pnpm test` — 6 files, 43 tests green.
- `pnpm build` — succeeds.

Fixes #31

### DIFF
--- a/server/api/put/rides/[id].ts
+++ b/server/api/put/rides/[id].ts
@@ -1,10 +1,30 @@
+import { z } from 'zod'
 import { prisma } from '../../../utils/prisma'
 import { sendEmail } from '../../../utils/email'
 import { sendNotification } from '../../../utils/notification'
+import { readValidatedBody } from '../../../utils/validation'
+
+// Whitelist of the ONLY fields an update may set (issue #31). Anything else
+// (clientId, createdAt, address ids, unknown keys, …) is rejected with a 400
+// rather than mass-assigned into prisma.ride.update. All fields are optional so
+// partial updates (e.g. the "mark complete" flow sending only status +
+// totalRideTime) still work; unknown keys are rejected via .strict().
+const updateRideSchema = z
+  .object({
+    status: z.enum(['CREATED', 'ASSIGNED', 'COMPLETED']).optional(),
+    // Empty string means "unassign"; handled below into null.
+    volunteerId: z.string().nullable().optional(),
+    scheduledTime: z.string().min(1).optional(),
+    pickupTime: z.string().min(1).nullable().optional(),
+    totalRideTime: z.number().optional(),
+    notes: z.string().nullable().optional(),
+    pickupDisplay: z.string().min(1).optional(),
+    dropoffDisplay: z.string().min(1).optional(),
+  })
+  .strict()
 
 export default defineEventHandler(async (event) => {
   const id = getRouterParam(event, 'id')
-  const body = await readBody(event)
 
   if (!id) {
     throw createError({
@@ -12,6 +32,8 @@ export default defineEventHandler(async (event) => {
       statusMessage: 'ID is required',
     })
   }
+
+  const body = await readValidatedBody(event, updateRideSchema)
 
   // Fetch existing ride to check previous state
   const existingRide = await prisma.ride.findUnique({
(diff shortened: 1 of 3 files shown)


TusharW4ni:

| Gate | Result |
|---|---|
| Independent reviewer suite | ✅ 43 real tests pass, 0 assertion failures (one file hit a 120s cold-boot hook timeout under load; passed in isolation — harness artifact, not this PR) |
| Revert-check | ✅ mass-assignment + unknown-key cases return 500 without the fix, 400 with it |
| Frontend cross-check | ✅ all 7 edit-form fields + the mark-complete `{status,totalRideTime}` payload are whitelisted — no legitimate UI update regresses |
| Scope | ✅ endpoint + new validation helper + test, ~57 net non-test lines |
| Risk files | ✅ none |

Deferred to follow-up (noted in PR): `post/rides`, `put/clients`, `put/volunteers` build their Prisma data explicitly (no mass-assignment); the new `readValidatedBody` helper makes them a drop-in later.

PR #46, opened 2026-07-06
Fix #21: derive trusted origins from environment

### LINKED ISSUE #21: [LOW] Hardcoded Trusted Origins in Authentication Settings
### Location
`server/utils/auth.ts`

### Description
The `better-auth` configuration has `trustedOrigins: ['http://localhost:3000', 'http://192.168.4.240:3000']` hardcoded. If deployed to production, login attempts from the actual production domain will be blocked due to CSRF/Origin validation failures.

### Recommended Fix
Rely on `process.env.APP_URL` or dynamic headers to populate the trusted origins dynamically.

TusharW4ni:
## What was broken
`server/utils/auth.ts` hardcoded `trustedOrigins: ['http://localhost:3000', 'http://192.168.4.240:3000']`. In production, better-auth's CSRF/Origin validation would reject logins from the real production domain (it isn't in the list), while a stale LAN IP was trusted.

## What changed
- Added `resolveTrustedOrigins(env)` — a pure, env-injected helper in `server/utils/auth.ts` that builds the trusted-origins list from:
  - `BETTER_AUTH_URL` and/or `APP_URL` (env vars already used elsewhere in this project),
  - an optional comma-separated `TRUSTED_ORIGINS` for extra origins (trimmed; blanks ignored),
  - always including `http://localhost:3000` so local dev and the e2e harness keep working.
  - Duplicates are de-duplicated.
- Dropped the hardcoded `192.168.4.240` LAN IP (add it back via `TRUSTED_ORIGINS` only if needed).
- Passed the helper result into `betterAuth({ trustedOrigins: resolveTrustedOrigins(process.env) })`.
- Documented the optional `TRUSTED_ORIGINS` var in `.env.example`.

No other auth behavior (OTP, hooks, session, rate limit) was touched.

## Verification
- New regression test `tests/e2e/trusted-origins.test.ts` (`describe('resolveTrustedOrigins')`) — a pure-function unit test that does **not** call `setup()`, so it does not boot the Nuxt app and stays fast. It asserts: no-env → includes `http://localhost:3000` and **not** the old `http://192.168.4.240:3000`; `BETTER_AUTH_URL`/`APP_URL` set → that origin is included; `TRUSTED_ORIGINS="https://a.com,https://b.com"` → both included; whitespace trimmed / blanks ignored; no duplicates.
- **Pre-fix**: the new test failed with `TypeError: resolveTrustedOrigins is not a function` (helper did not exist / origins were hardcoded). **Post-fix**: passes.
- **Full suite green**: `pnpm test` → `Test Files 7 passed (7)`, `Tests 51 passed (51)`. This confirms `loginAs()`/OTP login still works end-to-end against localhost with the new default.
- **Build**: `pnpm build` succeeds.

## NOTE — needs human review (auth config change)
This changes authentication configuration. For login to work in production after this change, the deployment **must set at least one** of:
- `BETTER_AUTH_URL` = the production origin (e.g. `https://rides.example.org`), and/or
- `APP_URL` = the production origin.

Optionally set `TRUSTED_ORIGINS` (comma-separated) for any additional trusted origins. If neither `BETTER_AUTH_URL` nor `APP_URL` is set in production, only `http://localhost:3000` is trusted and production logins will fail — so this env config is required at deploy time.

Fixes #21

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,9 @@
 DATABASE_URL=file:./dev.db
 BETTER_AUTH_SECRET=
 BETTER_AUTH_URL=http://localhost:3000
+# Optional: extra comma-separated origins trusted for CSRF/Origin validation.
+# BETTER_AUTH_URL (and APP_URL, if set) are trusted automatically.
+# TRUSTED_ORIGINS=https://rides.example.org
 EMAIL_USER=
 EMAIL_PASS=""
 NUXT_PUBLIC_GOOGLE_MAPS_API_KEY=
--- a/server/utils/auth.ts
+++ b/server/utils/auth.ts
@@ -6,6 +6,40 @@ import nodemailer from 'nodemailer'
 import { createAuthMiddleware, APIError } from 'better-auth/api'
 import { sendEmail } from './email'
 
+/**
+ * Resolve the list of origins better-auth trusts for CSRF/Origin validation.
+ *
+ * Derives the list from the environment instead of hardcoding a value, so a
+ * production deployment trusts its real domain (issue #21):
+ *   - BETTER_AUTH_URL / APP_URL — the app's own origin(s), already used
+ *     elsewhere in this project.
+ *   - TRUSTED_ORIGINS — optional comma-separated list of extra origins.
+ * `http://localhost:3000` is always included so local dev and the e2e harness
+ * keep working.
+ *
+ * Pure and env-injected so it can be unit-tested without booting better-auth.
+ */
+export function resolveTrustedOrigins(
+  env: Record<string, string | undefined>,
+): string[] {
+  const origins = ['http://localhost:3000']
+
+  for (const key of ['BETTER_AUTH_URL', 'APP_URL']) {
+    const value = env[key]?.trim()
+    if (value) origins.push(value)
+  }
+
+  const extra = env.TRUSTED_ORIGINS
+  if (extra) {
+    for (const origin of extra.split(',')) {
+      const trimmed = origin.trim()
+      if (trimmed) origins.push(trimmed)
+    }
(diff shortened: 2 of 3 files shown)


TusharW4ni:
🤖 **Auto-review gate: FLAGGED for human review — auth-config change.**

The fix is verified sound, but the gate **always** routes auth-configuration changes to a human before they deploy (merging = production deploy). This is that review, not a defect.

| Check | Result |
|---|---|
| Full suite (author-run) | ✅ 51/51 across 7 files — `loginAs()`/OTP login still works with the localhost default |
| Helper unit test | ✅ pure `resolveTrustedOrigins(env)`, 8 assertions, no app boot; fails pre-fix (`is not a function`), passes post |
| Diff | ✅ +38 net; `server/utils/auth.ts` + `.env.example` only |
| Correctness | ✅ always includes localhost, derives from `BETTER_AUTH_URL`/`APP_URL` + optional `TRUSTED_ORIGINS`, de-duped; drops the stale LAN IP |

### ⚠️ ACTION REQUIRED before/at merge — production login will break otherwise
This change makes trusted origins **env-driven**. In the production environment you MUST set **at least one** of:
- `BETTER_AUTH_URL` = the real prod origin (e.g. `https://rides.example.org`), or
- `APP_URL` = same.
- (optional) `TRUSTED_ORIGINS` = comma-separated extra origins.

If neither is set, only `http://localhost:3000` is trusted and **all production logins fail CSRF/origin validation**. Confirm the deploy env has these before merging.

Note: independent-reviewer cycle was skipped here intentionally — an auth-config PR flags regardless of verdict, so the human review below is the authoritative check.

Issue #25, opened 2026-04-02
[MEDIUM] Swallowed SMTP Errors Lead to False-Positive UI Feedback


TusharW4ni:
### Location
`server/utils/email.ts` (line 48)

### Description
The `sendEmail` utility silently traps and logs `transporter.sendMail` errors with the comment `// Don't rethrow, just log so it doesn't crash`. While this is fine for background notifications, it breaks endpoints where email delivery is the *primary, user-requested action* (e.g., OTP delivery for login, or sending a test notification via the Admin panel). If SMTP fails or credentials expire, the API returns HTTP 200, and the frontend falsely displays "OTP sent successfully" or "Test email sent".

### Recommended Fix
Either rethrow errors so they can be caught by the parent `defineEventHandler`, or return a success/failure boolean from `sendEmail` so critical paths (like OTP generation) can return a 500 error to the user if the email fails to send.

TusharW4ni:
🤖 **Skipped by `/issue-loop` — needs a design decision that collides with the test harness.**

The fix (surface SMTP failures on user-facing paths instead of swallowing them) is risky to automate because the e2e harness's `loginAs()` **depends on OTP-send failures being swallowed**: better-auth stores the OTP in the `verification` table *before* the send, and the test reads it there; if a failed send started returning 500 on `/api/auth/email-otp/send-verification-otp`, the entire authenticated test suite would break.

**Decision needed from a human:**
1. Which paths should surface send failures to the user (e.g. the admin 'send test email' button — clearly yes) vs. stay fire-and-forget (ride notifications — probably no)?
2. For the OTP path specifically: in production, if SMTP is down the user can't receive the code anyway — do we want a 500 there, and if so, how should the test harness continue to authenticate (e.g. a test-only bypass)?

Once you decide (1)/(2) I can implement it cleanly. Not auto-fixing because a wrong call here silently breaks every auth-gated test.

TusharW4ni:
🧑‍⚖️ **DECISION (owner, this session):** Refactor `sendEmail` to **return a success/failure boolean** everywhere; each caller decides whether to surface it. Surface failures on the OTP-send + admin test-email paths; keep background ride notifications fire-and-forget. Add a test-only bypass so the OTP-login harness (which relies on swallowed sends) keeps working. Touches the auth OTP path → human review.

Issue #45, opened 2026-07-06
Test infrastructure: per-file rebuild/boot is slow, and no way to test concurrency (blocks #47)


TusharW4ni:
This issue tracks the e2e harness improvements needed to keep the automated review/merge loop reliable. Two related problems, both surfaced by the loop.

## Problem 1 — per-file rebuild/boot is slow and flaky (surfaced by #44)

Each `tests/e2e/*.test.ts` calls `await setup()` (@nuxt/test-utils), which builds and boots the app independently. With the suite now ~9 files, a full `pnpm test` boots the app ~9× and, on a loaded machine (or a fresh worktree also running `pnpm install` + `prisma generate`), the first `setup()` cold build can exceed **@nuxt/test-utils' default 120s beforeAll timeout** — a spurious suite failure that passes on isolated re-run.

This is a reliability problem: the auto-merge gate requires a green full suite, and flaky boot-timeouts force manual re-runs.

### Options
- Stopgap: raise the `setup()` timeout (its own option, separate from vitest `hookTimeout`).
- Better: build the app once and share a single running server across all e2e files (one global setup boots the server; tests hit it over HTTP) instead of per-file `setup()`.
- `--no-file-parallelism` is already set; the cost is the repeated build/boot, not parallelism.

## Problem 2 — no way to write a failing test for concurrency bugs (blocks #47)

**PR #47 (fixes #12, the ride-signup TOCTOU race) is currently flagged and cannot auto-merge because its regression test passes with *and* without the fix.** The test harness uses the **synchronous better-sqlite3 adapter**, which serializes all writes, so the check-then-act race window physically cannot open through HTTP in-process. The committed test therefore pins the *invariant* (one winner, no overwrite) but does **not** fail on the un-fixed code — so it provides no CI protection against the race being reintroduced, which is why the gate holds it for human review.

During review of #47 the race was only reproducible by artificially neutralizing the in-memory pre-check and injecting a delay between read and write — a contrived, non-committable hack.

The harness needs a supported way to write regression tests for concurrency / atomicity bugs that genuinely fail pre-fix. Options to evaluate:
- Run e2e tests against an adapter/DB that does **not** fully serialize writes (e.g. a Postgres or libSQL/Turso test backend, or better-sqlite3 in a mode that allows interleaving), so real concurrent HTTP requests can interleave.
[...]

PR #47, opened 2026-07-06
Fix #12: atomic ride signup to close signup race

### LINKED ISSUE #12: [HIGH] Race Condition in Ride Signup (Time-of-Check to Time-of-Use)
### Location
`server/api/post/rides/[id]/signup.ts`

### Description
The endpoint checks if `ride.status === 'CREATED'` in memory, then runs a separate `update` query to set `volunteerId` and `status: 'ASSIGNED'`. Two volunteers clicking "Sign Up" simultaneously will both pass the in-memory check, and the second one will silently overwrite the first one's assignment.

### Recommended Fix
Use atomic updates by including `status: 'CREATED'` and `volunteerId: null` strictly inside the `where` clause of the Prisma `update` command.

TusharW4ni:
## What was broken

`server/api/post/rides/[id]/signup.ts` performed a classic check-then-act (TOCTOU) race: it read the ride with `findUnique`, checked `ride.status === 'CREATED'` in memory, then issued a **separate** `prisma.ride.update({ where: { id }, ... })` setting `volunteerId` + `status: 'ASSIGNED'`. Two volunteers signing up at the same time could both pass the in-memory check, and the second update would silently overwrite the first — the losing volunteer's assignment vanishes with no error.

## What changed

- **`signup.ts`** — moved the precondition into the update's WHERE clause: `where: { id, status: 'CREATED', volunteerId: null }`. The assignment now only succeeds while the ride is still unclaimed. If the ride was taken in the meantime, Prisma throws `P2025` (record to update not found); we catch it and return a clean **409 "Ride is no longer available"** instead of overwriting. The existing pre-checks (volunteer exists, AVAILABLE, ride exists/CREATED) are kept for good error messages, but the WHERE-clause guard is the real fix. Winner-only notification side effects are preserved — they run only after a successful guarded update, so a lost race sends no "you signed up" email.
- **`unsignup.ts`** — same check-then-act shape (lower stakes), given the identical atomic guard (`where: { id, status: 'ASSIGNED', volunteerId: volunteer.id }` + P2025 → 409). Small and clearly-correct; kept in this PR per the issue's verification comment.

No schema/auth/CI/docker/deps changes.

## Verification

- **Regression test:** `tests/e2e/signup-race.test.ts`
  - `two concurrent signups for the same ride: exactly one wins, no overwrite` — fires two genuinely concurrent signups (bob + alice, both seeded AVAILABLE) via `Promise.allSettled` of raw signup requests, each with its own `loginAs` cookie. Asserts exactly one 200, exactly one 409/400, and that the ride ends up ASSIGNED to whichever request returned 200 (never overwritten).
  - `signing up for an already-ASSIGNED ride is rejected (deterministic guard)` — bob claims the ride, then alice's attempt is rejected and bob's assignment is unchanged.
- `pnpm test` — full suite green: **8 files, 53 tests passed** (SMTP "Missing credentials" logs are the expected swallowed email failures, issue #25).
- `pnpm build` — succeeds.

### Note on forcing the race pre-fix
The test DB uses the synchronous **better-sqlite3** adapter, which serializes writes and does not yield mid-handler. Through the HTTP layer the pre-fix code therefore never actually double-assigns — even under a 20-way concurrent probe it consistently produced exactly `1×200, 19×400` (the loser's separate read already sees ASSIGNED and the in-memory pre-check returns 400). The regression tests consequently pin the **invariant** (exactly one winner, assignment never overwritten), which the WHERE-clause guard guarantees under any adapter that does interleave. The fix itself is the canonical, verified solution named in the issue's 2026-07-04 verification comment, confirmed present by the diff.

Fixes #12

### DIFF
--- a/server/api/post/rides/[id]/signup.ts
+++ b/server/api/post/rides/[id]/signup.ts
@@ -1,3 +1,4 @@
+import { Prisma } from '../../../../../prisma/generated/client'
 import { prisma } from '../../../../utils/prisma'
 import { auth } from '../../../../utils/auth'
 import { sendEmail } from '../../../../utils/email'
@@ -71,14 +72,35 @@ export default defineEventHandler(async (event) => {
     })
   }
 
-  // 3. Assign Volunteer
-  const updatedRide = await prisma.ride.update({
-    where: { id },
-    data: {
-      volunteerId: volunteer.id,
-      status: 'ASSIGNED'
+  // 3. Assign Volunteer (atomic — issue #12)
+  // The pre-checks above give good error messages, but they are a read
+  // separate from the write below: two volunteers can both pass the in-memory
+  // `ride.status === 'CREATED'` check and then race to update, with the second
+  // silently overwriting the first (TOCTOU). Put the precondition inside the
+  // WHERE clause so the assignment only succeeds while the ride is still
+  // unclaimed. If it was taken in between, Prisma throws P2025 (record to
+  // update not found) and we reject cleanly instead of overwriting.
+  let updatedRide
+  try {
+    updatedRide = await prisma.ride.update({
+      where: { id, status: 'CREATED', volunteerId: null },
+      data: {
+        volunteerId: volunteer.id,
+        status: 'ASSIGNED'
+      }
+    })
+  } catch (err) {
+    if (
+      err instanceof Prisma.PrismaClientKnownRequestError &&
+      err.code === 'P2025'
+    ) {
+      throw createError({
+        statusCode: 409,
+        statusMessage: 'Ride is no longer available',
+      })
     }
-  })
+    throw err
+  }
 
   // 4. Notifications
   const admins = await prisma.user.findMany({
--- a/server/api/post/rides/[id]/unsignup.ts
+++ b/server/api/post/rides/[id]/unsignup.ts
@@ -1,3 +1,4 @@
+import { Prisma } from '../../../../../prisma/generated/client'
 import { prisma } from '../../../../utils/prisma'
 import { auth } from '../../../../utils/auth'
 import { sendEmail } from '../../../../utils/email'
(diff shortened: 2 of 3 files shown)


TusharW4ni:
🤖 **Auto-review gate: FLAGGED — fix proven correct, but the committed test can't pin it (revert-check inconclusive). One-click merge recommended.**

This is **not** a defect in the fix. The independent reviewer verified the fix is correct — and went further than the standard check to prove it:

| Check | Result |
|---|---|
| Full suite (reviewer-run) | ✅ 53/53 green |
| Fix correctness | ✅ atomic `where:{id,status:'CREATED',volunteerId:null}` + P2025→409; verified by inspection |
| **Empirical race proof** | ✅ reviewer neutralized the in-memory pre-check + injected a 20ms delay on **un-fixed** code → reproduced the double-assign (`200,200`); with the atomic guard restored → exactly **one** winner |
| Strict improvement | ✅ can't break any working flow; frontend handles the new 409 generically |
| Scope / risk files | ✅ signup.ts + unsignup.ts + test, +30 net; no risk files |
| **Revert-check** | ⚠️ **INCONCLUSIVE** — both tests pass *without* the fix |

### Why flagged despite being proven correct
The synchronous better-sqlite3 test adapter serializes writes, so the TOCTOU window can't open through HTTP in the harness. The committed `signup-race.test.ts` therefore pins the *invariant* (one winner, no overwrite) but **passes with or without the fix** — meaning it won't catch a future regression that removes the guard. The auto-merge gate treats "no CI protection against reintroduction" as a bar to auto-**deploy** (merge = prod deploy here), independent of how confident we are today.

**Recommended action:** merge it — the fix is the canonical fix named in the issue and is empirically proven. The residual gap is test infrastructure, tracked by **#45** (an interleaving-capable adapter or DB-level concurrency test would let a future version of this test fail pre-fix and auto-merge cleanly).

Issue #33, opened 2026-07-04
Prisma Migrations Are Gitignored — No Reproducible Schema History


TusharW4ni:
### Location
`.gitignore` (`/prisma/migrations`), `package.json` (`prisma:reset` script)

### Description
`prisma/migrations` is gitignored, so the schema's change history exists only on individual dev machines. Consequences:

- **No reproducible schema in prod.** There is no way to run `prisma migrate deploy` against the production database; schema changes have to be applied by hand (which is how a raw `prod.sql` dump ended up committed — see the PII issue).
- **Drift between teammates.** Each dev's local migration folder diverges; `migrate dev` on one machine can generate migrations that conflict with another's.
- The only reset path is the destructive `prisma:reset` script (`migrate dev --name init && migrate reset -f && db seed`), which drops all data — fine for dev, unusable for prod.

### Recommended Fix
1. Remove `/prisma/migrations` from `.gitignore`.
2. Squash to a clean baseline: delete local migration folders, run `prisma migrate dev --name init` once against an empty dev DB, and commit the generated folder.
3. Mark the baseline as applied in production with `prisma migrate resolve --applied <init>`, then use `prisma migrate deploy` in the Docker entrypoint or CI for all future schema changes (relevant for #5, #23, #27 — all of which need migrations).

TusharW4ni:
🤖 **Skipped by `/issue-loop` — deployment-process decision, not safe to automate.**

Un-gitignoring `prisma/migrations` and establishing a baseline must coordinate with the **already-deployed production database**: after committing a squashed `init` migration you have to mark it as already-applied in prod (`prisma migrate resolve --applied <init>`) or `migrate deploy` will try to re-run it against a live DB that already has the schema. It also changes the team's dev workflow (from `db push` to `migrate dev`) and touches the Docker entrypoint/CI to run `migrate deploy`.

**Needs a human to:** confirm the prod DB's current state, choose the baseline commit, and decide where `migrate deploy` runs (Docker CMD vs CI). Happy to prepare the migration scaffolding + entrypoint change as a PR once you confirm the prod baseline — but the resolve/deploy step against production is yours.

TusharW4ni:
[...]

Issue #27, opened 2026-04-02
[Feature/Architecture] Soft Deletes for Auditability


TusharW4ni:
### Description
Currently, deleting a client, volunteer, or ride permanently removes the record from the database (`.delete()`). This destroys historical metrics and prevents recovery of accidental deletions.

### Implementation Details
- Update the Prisma schema to add a `deletedAt DateTime?` field to core models (Ride, User, Client, Volunteer).
- Change `DELETE` API endpoints to perform an `update` setting `deletedAt = new Date()`.
- Update all `GET` queries to filter with `where: { deletedAt: null }`.
- This ensures historical records remain intact while hiding them from active UI views.

Issue #8, opened 2026-04-02
[MEDIUM] Orphaned Users Due to Incomplete Deletion Cascades


TusharW4ni:
### Location
`server/api/delete/volunteers/[id].ts` and `server/api/delete/clients/[id].ts`

### Description
These endpoints delete the `Volunteer` or `Client` profile records but leave the parent `User` record completely intact. This leaves an orphaned User possessing a `VOLUNTEER` or `CLIENT` role but no profile, which will cause frontend queries (like viewing people) to crash when checking `user.volunteer.id` or similar mappings.

### Recommended Fix
Consider deleting the underlying `User` record if their singular client/volunteer profile is deleted, or safely downgrade their role.

Issue #7, opened 2026-04-02
[MEDIUM] Unassigning a Volunteer Leaves Ride "ASSIGNED"


TusharW4ni:
### Location
`server/api/put/rides/[id].ts`

### Description
If an admin updates a ride and clears the `volunteerId` parameter to an empty string (`""`), the code sets `updateData.volunteerId = null`. However, the status update is entirely ignored, meaning the ride remains permanently stuck in the `ASSIGNED` status with no volunteer.

### Recommended Fix
Automatically update the `status` to `CREATED` whenever `volunteerId` is set to `null`.

TusharW4ni:
**Verified 2026-07-04** — still present in `server/api/put/rides/[id].ts` (volunteerId cleared without touching `status`). Note that `server/api/post/rides/[id]/unsignup.ts` already does this correctly (`volunteerId: null, status: 'CREATED'` in one update) — the fix is to mirror that pattern.

PR #48, opened 2026-07-06
Fix #23: don't crash (or lose data) when deleting a client with rides

### LINKED ISSUE #23: Foreign Key Constraint Crashes on Client Deletion
### Location
`server/api/delete/clients/[id].ts` and `server/api/delete/volunteers/[id].ts`
`prisma/schema.prisma`

### Description
Attempting to delete a `Client` or `Volunteer` profile via the Admin dashboard will immediately crash the server with a `P2003 Foreign key constraint failed` exception if the person has *any* associated rides. This occurs because the `Ride` model defines relationships (`client` and `volunteer`) without specifying `onDelete: Cascade` or `onDelete: SetNull`. 

### Recommended Fix
Update `prisma/schema.prisma` to include explicit delete behaviors. For example:
`client Client @relation(fields: [clientId], references: [id], onDelete: Cascade)`
`volunteer Volunteer? @relation(fields: [volunteerId], references: [id], onDelete: SetNull)`

TusharW4ni:
## What was broken

The `ride -> client` foreign key is `ON DELETE RESTRICT` (confirmed in the verification comment on #23). Deleting a `Client` who had any rides threw `P2003 Foreign key constraint failed`, which surfaced as a **500** and made the admin UI delete fail. Separately, deleting a `Volunteer` didn't crash (the optional `volunteerId` FK is `ON DELETE SET NULL`), but their rides silently stayed `status: ASSIGNED` with a null volunteer — stuck, never returning to the available pool (the #7 stuck state).

## What changed (endpoint logic only — no schema change)

- `server/api/delete/clients/[id].ts`: count the client's rides first; if any exist, return a clear **409** (`Cannot delete a client with existing rides — reassign or archive them first`) instead of crashing. Ride-less clients still delete normally. A `P2003` catch wraps the delete as a race-safe fallback (a ride created between the count and the delete still yields a 409, never a 500).
- `server/api/delete/volunteers/[id].ts`: in a `$transaction`, reset the volunteer's `ASSIGNED` rides back to `CREATED` before deleting the volunteer, so those rides return to the available pool instead of getting stuck.

### Why block instead of cascade
This is the Data-Integrity milestone. Cascade-deleting a client's rides would wipe `COMPLETED`-ride history that the metrics endpoints depend on. A non-destructive 409 block preserves history and is trivially reversible once soft-delete (#27) lands, which would supersede this. **No Prisma schema change** (which would also need a migration, blocked on #33).

## Verification

New regression test `tests/e2e/safe-delete.test.ts` (`safe deletes (#23)`):
- *does not crash (500) when deleting a client with rides — blocks with 409 and preserves data* — deletes seeded `martha` (has rides); pre-fix returned **500** (P2003), now **409**, and the client + all their rides still exist afterward.
- *still deletes a client that has no rides (200)* — creates a fresh client via `POST /api/post/clients`, deletes it, expects **200**.
- *resets a deleted volunteer's ASSIGNED rides back to CREATED (not stuck)* — creates a volunteer, assigns them a ride (status `ASSIGNED`), deletes the volunteer; pre-fix the ride stayed `ASSIGNED`, now it's `CREATED` with a null volunteer.

Confirmed the two failing assertions pre-fix (500 for client-with-rides; `ASSIGNED` for the volunteer's ride). Full suite green (`pnpm test`: 54 passed) and `pnpm build` succeeds.

## Related (not resolved here)
Overlaps with #7 (stuck ASSIGNED rides) and #27 (soft deletes), and #8 (orphaned users — a deleted client/volunteer's `User` row remains); kept out of scope.

Fixes #23

### DIFF
--- a/server/api/delete/clients/[id].ts
+++ b/server/api/delete/clients/[id].ts
@@ -1,3 +1,4 @@
+import { Prisma } from '../../../../prisma/generated/client'
 import { prisma } from '../../../utils/prisma'
 
 export default defineEventHandler(async (event) => {
@@ -10,7 +11,37 @@ export default defineEventHandler(async (event) => {
     })
   }
 
-  return await prisma.client.delete({
-    where: { id },
-  })
+  // Issue #23: the ride->client FK is ON DELETE RESTRICT, so deleting a client
+  // who has any rides throws P2003 and used to surface as a 500. We must not
+  // crash, and we must not destroy historical ride data (COMPLETED rides feed
+  // the metrics endpoints — this is the Data-Integrity milestone). Block the
+  // delete with a clear 409 while rides exist; only ride-less clients delete.
+  const rideCount = await prisma.ride.count({ where: { clientId: id } })
+  if (rideCount > 0) {
+    throw createError({
+      statusCode: 409,
+      statusMessage:
+        'Cannot delete a client with existing rides — reassign or archive them first',
+    })
+  }
+
+  try {
+    return await prisma.client.delete({
+      where: { id },
+    })
+  } catch (err) {
+    // Race-safe fallback: a ride could be created between the count and the
+    // delete. Treat the FK violation as the same 409 block, never a 500.
+    if (
+      err instanceof Prisma.PrismaClientKnownRequestError &&
+      err.code === 'P2003'
+    ) {
+      throw createError({
+        statusCode: 409,
+        statusMessage:
+          'Cannot delete a client with existing rides — reassign or archive them first',
+      })
+    }
+    throw err
+  }
 })
--- a/server/api/delete/volunteers/[id].ts
+++ b/server/api/delete/volunteers/[id].ts
@@ -10,7 +10,20 @@ export default defineEventHandler(async (event) => {
     })
   }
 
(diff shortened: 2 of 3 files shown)


TusharW4ni:

| Gate | Result |
|---|---|
| Independent reviewer suite | ✅ 54/54 green (clean-worktree re-run) |
| Revert-check | ✅ client-with-rides returns 500 without the fix, 409 with it; volunteer ride stuck ASSIGNED without, reset to CREATED with |
| Scope | ✅ 2 delete endpoints + test, net +54 |
| Risk files | ✅ none (endpoint logic only, no schema change — avoids the #33 migration block) |
| `/review` findings | ✅ P2003 caught precisely, other errors rethrown (no false 409); volunteer reset is transactional; frontend shows an error toast on 409 |
| Priority | P1 data-integrity (not P0 security) → eligible for auto-merge |

Non-blocking notes carried forward: (1) the delete-failure toast is generic rather than surfacing the 'reassign or archive' message; (2) a client with only COMPLETED rides can't be deleted until soft-delete (#27) lands — the correct tradeoff for this milestone. Overlaps #7/#8/#27, kept out of scope.

PR #50, opened 2026-07-07
Fix #7: reset ride to CREATED when volunteer is unassigned

### LINKED ISSUE #7: [MEDIUM] Unassigning a Volunteer Leaves Ride "ASSIGNED"

TusharW4ni:
## What was broken

When an admin edited a ride and cleared the volunteer (`volunteerId` empty string → `updateData.volunteerId = null`), the endpoint left `status` untouched. The ride stayed **ASSIGNED with no volunteer** — a stuck state where the ride never returned to the available pool. Confirmed present on `main`; the verification comment (2026-07-04) noted `server/api/post/rides/[id]/unsignup.ts` already handles this correctly (`volunteerId: null, status: 'CREATED'` in one update) and that the fix is to mirror that pattern.

## What changed

`server/api/put/rides/[id].ts` — in the volunteer assignment/unassignment block, when the update resolves `volunteerId` to `null` **and** the caller did not send an explicit `status`, set `status = 'CREATED'`. If the caller sent an explicit status (e.g. a deliberate `COMPLETED`), it is respected and not overridden. Validation schema, notifications, and date handling are unchanged. Net +9 lines in the endpoint.

## Verification

New regression suite `tests/e2e/unassign-status.test.ts`:

- **resets an ASSIGNED ride to CREATED when the volunteer is cleared** — PUT `{ volunteerId: '' }` on an ASSIGNED ride; asserts `status: 'CREATED'` and `volunteerId: null`. This is the assertion that **failed pre-fix** (`expected 'ASSIGNED' to be 'CREATED'`).
- **respects an explicit status when unassigning (does not force CREATED)** — PUT `{ volunteerId: '', status: 'COMPLETED' }`; asserts the explicit `COMPLETED` wins.
- **leaves status untouched on an edit that does not change the volunteer** — PUT `{ notes: ... }`; asserts status stays `ASSIGNED` (no accidental reset).
- **assigning a volunteer still works** — PUT `{ volunteerId, status: 'ASSIGNED' }` on a CREATED ride.

Confirmed the core test fails before the fix and passes after. Full suite green (61 passed / 11 files) and `pnpm build` succeeds. Tests restore mutated rides to seed state in `afterEach` so they don't disturb the shared-DB `pii-scoping` suite.

Fixes #7

### DIFF
--- a/server/api/put/rides/[id].ts
+++ b/server/api/put/rides/[id].ts
@@ -85,6 +85,15 @@ export default defineEventHandler(async (event) => {
   if (body.volunteerId !== undefined) {
     updateData.volunteerId =
       body.volunteerId && body.volunteerId.trim() !== '' ? body.volunteerId : null
+
+    // Issue #7: unassigning a volunteer must also return the ride to the
+    // available pool, otherwise it is left ASSIGNED with no volunteer — a stuck
+    // state. Mirror the unsignup endpoint (volunteerId: null, status: 'CREATED').
+    // Only auto-set CREATED when the caller didn't explicitly request a status
+    // (e.g. a deliberate COMPLETED must be respected).
+    if (updateData.volunteerId === null && body.status === undefined) {
+      updateData.status = 'CREATED'
+    }
   }
 
   const ride = await prisma.ride.update({
--- a/server/api/put/rides/[id].ts
+++ b/server/api/put/rides/[id].ts
@@ -89,9 +89,15 @@ export default defineEventHandler(async (event) => {
     // Issue #7: unassigning a volunteer must also return the ride to the
     // available pool, otherwise it is left ASSIGNED with no volunteer — a stuck
     // state. Mirror the unsignup endpoint (volunteerId: null, status: 'CREATED').
-    // Only auto-set CREATED when the caller didn't explicitly request a status
-    // (e.g. a deliberate COMPLETED must be respected).
-    if (updateData.volunteerId === null && body.status === undefined) {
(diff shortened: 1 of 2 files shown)


TusharW4ni:
🤖 **Auto-review gate: PASSED (after a hardening commit) — merging.**

| Gate | Result |
|---|---|
| Independent reviewer suite | ✅ 61/61 green (run twice, order-independent) |
| Revert-check | ✅ core test fails without the fix (status stays ASSIGNED), passes with it |
| Cross-test contamination | ✅ verified clean — afterEach restores touched rides; pii-scoping passes after unassign-status |
| Scope | ✅ 1 endpoint + test, ~+9 net |
| Risk files | ✅ none |

**Hardening applied before merge (commit ec3060d):** `/review` flagged that clearing a volunteer on a **COMPLETED** ride (the edit form never sends `status`) would silently flip it to CREATED — dropping it from completion metrics while leaving its `totalRideTime`. In the data-integrity milestone that's a real regression the fix would introduce, so I narrowed the auto-reset guard to `existingRide.status === 'ASSIGNED'` and added a regression test that a COMPLETED ride stays COMPLETED when unassigned. Full suite re-verified: 62/62 green.

PR #51, opened 2026-07-07
Fix #8: delete the User when its client/volunteer profile is removed

### LINKED ISSUE #8: [MEDIUM] Orphaned Users Due to Incomplete Deletion Cascades

TusharW4ni:
## What was broken

Deleting a `Client` or `Volunteer` profile removed only the profile row and left the parent `User` intact — an orphaned User with role `CLIENT`/`VOLUNTEER` but no profile. Frontend queries that read `user.client.id` / `user.volunteer.id` crash on such orphans.

## What changed

Both delete endpoints now delete the underlying **User** instead of just the profile. Because `Client.user` and `Volunteer.user` are `onDelete: Cascade`, deleting the User removes the profile row too, leaving no orphan. No schema change.

- `server/api/delete/clients/[id].ts` — keep #23's "409 if the client has rides" pre-check and the FK-violation fallback; when deletion proceeds, `prisma.user.delete({ where: { id: client.userId } })` cascades the client away. Added a 404 for a missing client id (was a P2025 500).
- `server/api/delete/volunteers/[id].ts` — keep #23's transactional `ASSIGNED → CREATED` ride reset, then delete the User inside the same transaction (the `ride → volunteer` FK is `SET NULL`, so reset rides keep pointing at null). Added a 404 for a missing volunteer id.

**FK error code:** deleting the User cascades to the client row, which trips the `ride → client` RESTRICT FK in the race window between the ride-count check and the delete. I verified empirically (better-sqlite3 adapter) that this surfaces as **P2003**; the fallback now catches **P2003 or P2014** defensively so a driver quirk can't leak a 500. Everything else is rethrown.

**Dual-profile edge:** roles are singular in this app, so a User won't hold both a client and a volunteer profile. Deleting the User is therefore safe; I proceed with User deletion and note the assumption in code comments rather than adding special-case handling.

## Verification

New regression test `tests/e2e/orphaned-users.test.ts`:
- `deleting a fresh volunteer removes its User (no orphan)`
- `deleting a fresh ride-less client removes its User (no orphan)`
- `deleting a client WITH rides still 409s and leaves client + user intact (#23)`
- `deleting a volunteer resets its ASSIGNED rides to CREATED AND removes the User (#23)`

Revert-check (source stashed, test kept) — orphan tests fail pre-fix with:
```
AssertionError: parent User should be deleted, not orphaned: expected { …(8) } to be undefined
+ Received: { "email": "orphan-vol-...@example.com", "role": "VOLUNTEER", ... }
```
With the fix restored, all pass.

- `pnpm test` — **66 passed (12 files)**, including the 4 new tests.
- `pnpm build` — succeeds.

Fixes #8

### DIFF
--- a/server/api/delete/clients/[id].ts
+++ b/server/api/delete/clients/[id].ts
@@ -25,16 +25,36 @@ export default defineEventHandler(async (event) => {
     })
   }
 
+  // Issue #8: deleting only the Client profile left an orphaned User (role
+  // CLIENT, no profile) that crashes frontend queries reading user.client.id.
+  // Delete the underlying User instead — Client.user is onDelete: Cascade, so
+  // this removes the client row too, leaving no orphan. (Roles are singular in
+  // this app, so a User won't also hold a volunteer profile.)
+  const client = await prisma.client.findUnique({
+    where: { id },
+    select: { userId: true },
+  })
+  if (!client) {
+    throw createError({
+      statusCode: 404,
+      statusMessage: 'Client not found',
+    })
+  }
+
   try {
-    return await prisma.client.delete({
-      where: { id },
+    return await prisma.user.delete({
+      where: { id: client.userId },
     })
   } catch (err) {
     // Race-safe fallback: a ride could be created between the count and the
-    // delete. Treat the FK violation as the same 409 block, never a 500.
+    // delete. We now delete the User, whose cascade to the client row is what
+    // trips the ride->client RESTRICT FK. Empirically (better-sqlite3 adapter)
+    // this surfaces as P2003; we also accept P2014 (required-relation
+    // violation) defensively so a driver quirk can't leak a 500. Treat both as
+    // the same 409 block.
     if (
       err instanceof Prisma.PrismaClientKnownRequestError &&
-      err.code === 'P2003'
+      (err.code === 'P2003' || err.code === 'P2014')
     ) {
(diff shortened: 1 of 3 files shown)


TusharW4ni:

| Gate | Result |
|---|---|
| Independent reviewer suite | ✅ 66/66 green |
| Revert-check | ✅ 3 orphan-User assertions fail without the fix, pass with it |
| #23 preservation | ✅ client-with-rides still 409s (data intact); volunteer ASSIGNED→CREATED reset intact |
| Scope | ✅ 2 delete endpoints + test, +45/-6 |
| Risk files | ✅ none |
| `/review` findings | ✅ FK cascade correctly caught as P2003/P2014→409, non-FK errors rethrow; dual-profile edge documented (roles singular, matches delete/admins convention) |

PR #52, opened 2026-07-07
Fix #15: store blank phone numbers as null to avoid unique-constraint crash

### LINKED ISSUE #15: [MEDIUM] Unique Constraint Violation on Empty Phone Numbers
### Location
`server/api/put/clients/[id].ts`, `server/api/put/volunteers/[id].ts`, `server/api/post/volunteers/index.ts`

### Description
The API accepts empty strings (`""`) from the frontend when a phone number is cleared. Because Prisma (with SQLite) treats `""` as a distinct value, attempting to clear the phone number for a second user will trigger a fatal `P2002 Unique constraint failed` error, crashing the request.

### Recommended Fix
Sanitize inputs to convert empty strings to `null` before passing them to Prisma for optional unique fields.

TusharW4ni:
## What was broken

The client/volunteer create and update endpoints wrote `body.phone` straight to `User.phone`. When the frontend clears a phone it sends `""`, and `User.phone` is `String? @unique`. SQLite treats `""` as a distinct value, so the first blank saves fine but clearing a **second** user's phone stores another `""` and trips the unique constraint — a `P2002` error that 500s the request.

## What changed

- Added `emptyToNull(value)` in `server/utils/sanitize.ts` (auto-imported by Nitro): normalizes empty / whitespace-only strings to `null`.
- Applied it to every site that writes `phone`:
  - `server/api/put/volunteers/[id].ts`
  - `server/api/put/clients/[id].ts`
  - `server/api/post/volunteers/index.ts` (create + update branches)
  - `server/api/post/clients/index.ts` (create + update branches)

No schema change — the `@unique` constraint stays; the fix is input sanitization, mirroring the existing `email: body.email || null` pattern.

## Verification

Regression test: `tests/e2e/empty-phone.test.ts`
- `clearing phone (phone: "") on two different volunteers both succeed and store null` — pre-fix the **second** clear failed with `500 Server Error` (`P2002 Unique constraint failed`); post-fix both succeed and the stored phone is `null`.
- `whitespace-only phone is also stored as null` — pre-fix stored `"   "`; post-fix `null`.
- Test restores seeded phone values in a `finally` block (shared DB).

Confirmed the test FAILS against unmodified source and PASSES after the fix.

- `pnpm test` — 13 files, **68 tests passed**.
- `pnpm build` — success.

Fixes #15

### DIFF
--- a/server/api/post/clients/index.ts
+++ b/server/api/post/clients/index.ts
@@ -23,7 +23,7 @@ export default defineEventHandler(async (event) => {
       data: {
         name: body.name,
         email: body.email || null,
-        phone: body.phone,
+        phone: emptyToNull(body.phone),
         role: 'CLIENT'
       }
     })
@@ -32,7 +32,7 @@ export default defineEventHandler(async (event) => {
       where: { id: user.id },
       data: {
         name: body.name,
-        phone: body.phone,
+        phone: emptyToNull(body.phone),
         // Don't downgrade ADMIN/VOLUNTEER
       }
     })
--- a/server/api/post/volunteers/index.ts
+++ b/server/api/post/volunteers/index.ts
@@ -23,7 +23,7 @@ export default defineEventHandler(async (event) => {
       data: {
         name: body.name,
         email: body.email,
-        phone: body.phone,
+        phone: emptyToNull(body.phone),
         role: 'VOLUNTEER'
       }
     })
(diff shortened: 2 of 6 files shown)


TusharW4ni:

| Gate | Result |
|---|---|
| Independent reviewer suite | ✅ 68/68 green |
| Revert-check | ✅ second phone-clear 500s (P2002) without the fix; whitespace stored literally; both pass with it |
| Scope | ✅ new helper + 6 call-sites across 4 endpoints, +20 net source |
| Risk files | ✅ none |
| `/review` findings | ✅ emptyToNull correct (blanks/whitespace/non-string→null, real phone preserved); duplicate real phones still rejected; all create+update branches covered |

Follow-up filed: the **admin** create/update endpoints carry the identical blank-phone bug (out of #15's scope).

PR #55, opened 2026-07-07
Fix #45: shared single test server + concurrency-testing capability

### LINKED ISSUE #45: Test infrastructure: per-file rebuild/boot is slow, and no way to test concurrency (blocks #47)

TusharW4ni:
Overhauls the e2e harness to remove the two problems that were slowing and blocking the review loop.

## Problem 1 — per-file rebuild/boot (slow + flaky timeout)
Each `tests/e2e/*.test.ts` called `@nuxt/test-utils` `setup()`, which built **and** booted the app independently — **12 server boots** per run. A full `pnpm test` took ~223s, and the first cold build could exceed `@nuxt/test-utils`' 120s per-file `beforeAll` timeout, producing spurious failures that forced manual re-runs.

**Fix:** `tests/global-setup.ts` now builds the app **once**, seeds the DB, and starts **one** server; every test file connects to it via a shared `bootShared()` helper (`setup({ host })`). The build runs inside `globalSetup`, which has no per-file hook timeout.

| | before | after |
|---|---|---|
| server boots | 12 | **1** |
| full suite (cold `pnpm test`) | ~223s | **~28s** |
| 120s per-file build timeout | possible | gone |

All 13 files / 68 tests pass (cold and warm).

## Problem 2 — concurrency was untestable (this is why #47's test couldn't fail pre-fix)
The synchronous better-sqlite3 adapter runs a handler's read and write too close together for a competing HTTP request to interleave, so a race test passed even against buggy code — #12/#47's regression test could only pin an invariant, never fail on the pre-fix code.

**Fix:** a **test-only** fault-injection seam, `server/utils/testHooks.ts` `raceDelay(event, label)`, that widens the read→write window. It is a strict **no-op in production** — it does nothing unless BOTH the server was started with `TEST_RACE_HOOKS=1` (set only by `tests/global-setup.ts`) AND the request carries a matching `x-test-race-delay: <label>:<ms>` header. Wired into the signup TOCTOU point.

`tests/e2e/signup-race.test.ts` now uses it and **genuinely fails on the pre-fix non-atomic code** — verified by reverting the guard to `where: { id }`: the concurrent test fails ("expected exactly one 200"), and passes again with the atomic guard restored. Future concurrency fixes can now satisfy the loop's revert-check honestly.

## Verification
- `pnpm test` — 13 files, 68 tests green (cold-cache run included).
- `pnpm build` — succeeds.
- Revert-check of the seam: guard removed → `signup-race` fails; guard restored → passes.

## Risk / scope
- Test infrastructure + one guarded no-op line in `signup.ts`. No schema/auth/CI/docker/dependency changes.
- The production behavior of `signup.ts` is unchanged when `TEST_RACE_HOOKS` is unset (i.e. always, outside the harness).

Fixes #45

### DIFF
--- a/server/api/post/rides/[id]/signup.ts
+++ b/server/api/post/rides/[id]/signup.ts
@@ -72,6 +72,10 @@ export default defineEventHandler(async (event) => {
     })
   }
 
+  // Test-only seam (issue #45): widen the read→write window so the signup race
+  // is reproducible through HTTP in e2e. No-op in production (see raceDelay).
+  await raceDelay(event, 'ride-signup')
+
   // 3. Assign Volunteer (atomic — issue #12)
   // The pre-checks above give good error messages, but they are a read
   // separate from the write below: two volunteers can both pass the in-memory
--- a/server/utils/testHooks.ts
+++ b/server/utils/testHooks.ts
@@ -1,0 +1,21 @@
+import type { H3Event } from 'h3'
+
+// Test-only fault-injection seam (issue #45).
+//
+// Concurrency bugs like the signup TOCTOU (#12) can't be reproduced through the
+// HTTP layer in the e2e harness: the synchronous better-sqlite3 adapter runs a
+// handler's read and write close enough together that a competing request never
+// interleaves between them, so a race test would pass even against buggy code.
+//
+// This seam lets a test deliberately widen that read→write window. It is a
+// complete no-op in production: it does nothing unless BOTH the server was
+// started with `TEST_RACE_HOOKS=1` (only tests/global-setup.ts sets this) AND
+// the request carries a matching `x-test-race-delay: <label>:<ms>` header.
+export async function raceDelay(event: H3Event, label: string): Promise<void> {
+  if (process.env.TEST_RACE_HOOKS !== '1') return
+  const header = getHeader(event, 'x-test-race-delay')
+  if (!header) return
+  const [target, ms] = header.split(':')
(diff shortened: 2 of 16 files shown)


TusharW4ni:
🤖 **Independent review: RECOMMEND_MERGE** (verified in a clean worktree by a reviewer with no author context).

| Check | Result |
|---|---|
| Full suite (warm) | ✅ 13 files / 68 tests, ~29s, **1** `Listening on` (single-server confirmed) |
| Cold run (CI path, caches cleared) | ✅ builds once in globalSetup, 1 boot, 68/68, ~38s, no timeout |
| 12-file transform integrity | ✅ only import + `bootShared()` swaps; no test body/assertion altered or lost (5 files read fully, 7 by diff) |
| Seam is a prod no-op | ✅ `raceDelay` short-circuits on `TEST_RACE_HOOKS !== '1'` before reading the header; `TEST_RACE_HOOKS` set only in global-setup; not abusable by a prod client |
| Seam pins the bug | ✅ reviewer reverted the atomic guard → `signup-race` FAILED (`200,200`); restored → passes 2/2 |
| Risk files | ✅ none; only prod change is one guarded no-op line in signup.ts |

Held for human merge (deploy discipline: merge = prod deploy, and it touches a production file). Recommended to merge — it unblocks the loop (≈8× faster cycles) and makes concurrency fixes auto-mergeable going forward.

PR #56, opened 2026-07-08
Fix #16: normalize addresses before upsert to prevent duplicates

### LINKED ISSUE #16: [MEDIUM] Case-Sensitive Address Duplication (Prisma Upsert)
### Location
`server/api/post/rides/index.ts` and `server/api/post/clients/index.ts`

### Description
The backend attempts to prevent duplicate addresses using `prisma.address.upsert` with a unique constraint on `[street, city, state, zip]`. However, it doesn't normalize casing. A user typing "123 Main st" versus "123 Main St" will spawn multiple identical but case-differing address records, causing database bloat.

### Recommended Fix
Normalize all address fields (e.g. `trim().toLowerCase()`) before executing the upsert.

TusharW4ni:
## What was broken
`prisma.address.upsert` on the `@@unique([street, city, state, zip])` key ran on the raw request body, so `"123 Main St"`, `"123 main st "` and `" 123 Main St "` each missed the constraint and spawned near-duplicate `Address` rows — bloating the table and defeating the constraint's intent. This affected all three upsert sites: ride-create (`resolveAddress`), client-create, and client-update.

## What changed
Added a shared, auto-imported `normalizeAddress()` helper in `server/utils/address.ts` and applied it before all three upserts. Normalization trims each field, collapses internal whitespace, **Title-Cases** street/city, upper-cases `state`, and trims `zip`.

### Casing choice + client-display tradeoff
The issue literally suggested `trim().toLowerCase()`, but the `Address` row backs `Client.homeAddress`, which is **rendered on the people page** — lowercasing would be a visible UI regression (`123 main st, dallas, tx`). I chose **trim + collapse whitespace + Title Case** instead: it dedups equivalent addresses onto one row while staying presentable. `Ride.pickupDisplay`/`dropoffDisplay` are separate snapshot strings built from the raw body, so they are unaffected either way.

## Verification
- **Regression test:** `tests/e2e/address-normalization.test.ts` → `address normalization on upsert (#16) > collapses case/whitespace-variant pickup addresses to a single Address row`. Creates two rides as admin with case/whitespace-variant pickup streets (`"<n> Norm Test Ave"` and `"  <n> norm test ave  "`) and asserts both resolve to the same `Address` id and the addresses search endpoint returns exactly one row.
- **Pre-fix failing assertion (observed):** `AssertionError: expected '0d199ed0-19e7-4486-bd6c-4a9fbc15b06d' to be '63f4d6b3-299c-4622-b0b1-3a244a440247'` at `expect(rideB.pickupAddressId).toBe(rideA.pickupAddressId)` — i.e. two distinct rows pre-fix.
- **Post-fix:** the test passes (one shared row).
- **Suite:** `pnpm test` — 14 files, 69 tests pass. `pnpm build` succeeds.

## Scope / risk
No schema, auth, CI, docker, or dependency changes. No risk files touched. Net +117 / -19 lines across 3 API routes + 1 new util + 1 new test.

Fixes #16

### DIFF
--- a/server/api/post/clients/index.ts
+++ b/server/api/post/clients/index.ts
@@ -39,15 +39,15 @@ export default defineEventHandler(async (event) => {
   }
 
   // 2. Upsert Address
-  // Using findFirst because unique constraint might not match perfectly if we don't normalize
-  // But strictly, we should use upsert or findUnique if strictly defined
-  // The schema has @@unique([street, city, state, zip])
-  const addressData = {
+  // Normalize first so case/whitespace-variant addresses collapse onto the
+  // @@unique([street, city, state, zip]) key instead of spawning duplicate
+  // rows (issue #16).
+  const addressData = normalizeAddress({
     street: body.street,
     city: body.city,
     state: body.state,
     zip: body.zip,
-  }
+  })
 
   const address = await prisma.address.upsert({
     where: {
--- a/server/api/post/rides/index.ts
+++ b/server/api/post/rides/index.ts
@@ -18,22 +18,15 @@ export default defineEventHandler(async (event) => {
     state: string
     zip: string
   }) => {
+    // Normalize before the upsert so case/whitespace-variant addresses collapse
+    // onto the @@unique([street, city, state, zip]) key (issue #16).
+    const normalized = normalizeAddress(addr)
     return await prisma.address.upsert({
       where: {
-        street_city_state_zip: {
-          street: addr.street,
-          city: addr.city,
-          state: addr.state,
-          zip: addr.zip,
-        },
+        street_city_state_zip: normalized,
       },
       update: {},
-      create: {
-        street: addr.street,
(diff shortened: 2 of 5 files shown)


TusharW4ni:

| Gate | Result |
|---|---|
| Independent reviewer suite | ✅ 69/69 green (~29s, shared-server harness) |
| Revert-check | ✅ two distinct Address ids without the fix, one shared id with it |
| Scope | ✅ shared helper + 3 upsert sites + test, +56 net |
| Risk files | ✅ none (no schema change; relies on existing unique constraint) |
| `/review` findings | ✅ normalization applied consistently (same value for where + create at all 3 sites); pickupDisplay snapshot unaffected; Title-Case cosmetic mangling of stored display judged an acceptable, documented tradeoff |

Follow-up filed for the proper canonical-key fix (needs a schema column, blocked on #33).

PR #58, opened 2026-07-08
Fix #18: include the full end day in metrics date range

### LINKED ISSUE #18: [LOW] End-of-Day Boundary Bug in Metrics Calculation
### Location
`server/api/get/metrics/completionRate/index.ts` and `server/api/get/metrics/hours/index.ts`

### Description
When filtering metrics by date, the frontend passes an `endDate` string (e.g., `2026-12-31`). `new Date()` parses this as UTC Midnight (`2026-12-31T00:00:00.000Z`). Because the Prisma query uses `lte: endDate`, any ride scheduled *during* the day on December 31st (e.g., 10:00 AM) is falsely excluded from the metrics.

### Recommended Fix
Modify the `endDate` boundary to include the full day (e.g. setting the time to 23:59:59).

TusharW4ni:
## What was broken

The dashboard metrics endpoints receive `endDate` as a plain `YYYY-MM-DD` day string from the frontend. `new Date('2026-12-31')` parses to UTC midnight, and the queries filtered with `lte: endDate` (`completionRate`, `hours`) or `lt: endDate` (`topRiders`), so any ride happening *during* the end day (e.g. 14:00) was wrongly excluded — the whole final day was silently dropped from the metrics. (issue #18; the verification comment flagged `topRiders` as having the same class of bug via `lt: endFilter`.)

## What changed

- New auto-imported helper `server/utils/dateRange.ts` → `parseDateRange(startDate, endDate)`, which returns a `{ gte, lt }` Prisma `scheduledTime` filter. Start stays start-of-day inclusive; the end boundary is bumped to **start of the next day** and compared with `lt`, so the entire end day is included. Returns `undefined` when both params are absent so callers keep their existing no-filter / default behavior.
- All three metrics endpoints (`completionRate`, `hours`, `topRiders`) now share this single implementation. `topRiders`' year-to-date default and its open-bound behavior (beginning-of-time / max-date when only one side is passed) are preserved.

**Timezone assumption (documented in the helper):** boundaries are computed in UTC to match how the plain `YYYY-MM-DD` strings are already parsed (`new Date('YYYY-MM-DD')` yields UTC midnight). This keeps parse and filter consistent.

## Verification

Regression test: `tests/e2e/metrics-date-range.test.ts`
- `counts an afternoon ride scheduled on the endDate day` — creates a COMPLETED ride at 14:00 on the `endDate` day, asserts the delta in `hours` and `completionRate` includes it.
- `includes the endDate day for topRiders (explicit-endDate path, lt bug)` — asserts the same ride appears in `topRiders` for a narrow explicit-endDate window.

**Pre-fix failing assertions (observed):**
- `expected +0 to be close to 3.25, received difference is 3.25, but expected 0.000005` (hours delta was 0 — the afternoon ride on `endDate` was excluded).
- `expected 0 to be greater than or equal to 1` (topRiders returned an empty list — `lt: endDate` dropped the whole end day).

Both pass after the fix. Full suite green (15 files / 71 tests); `pnpm build` succeeds.

Fixes #18

### DIFF
--- a/server/api/get/metrics/completionRate/index.ts
+++ b/server/api/get/metrics/completionRate/index.ts
@@ -6,15 +6,9 @@ export default defineEventHandler(async (event) => {
   requireAdmin(event)
 
   const query = getQuery(event)
-  const startDate = query.startDate ? new Date(String(query.startDate)) : undefined
-  const endDate = query.endDate ? new Date(String(query.endDate)) : undefined
-
-  const dateFilter = startDate || endDate ? {
-    scheduledTime: {
-      ...(startDate && { gte: startDate }),
-      ...(endDate && { lte: endDate })
-    }
-  } : {}
+  // Shared helper makes endDate inclusive of the whole day (issue #18).
+  const range = parseDateRange(query.startDate, query.endDate)
+  const dateFilter = range ? { scheduledTime: range } : {}
 
   const totalRides = await prisma.ride.count({
     where: dateFilter
--- a/server/api/get/metrics/hours/index.ts
+++ b/server/api/get/metrics/hours/index.ts
@@ -6,15 +6,9 @@ export default defineEventHandler(async (event) => {
   requireAdmin(event)
 
   const query = getQuery(event)
-  const startDate = query.startDate ? new Date(String(query.startDate)) : undefined
-  const endDate = query.endDate ? new Date(String(query.endDate)) : undefined
-
-  const dateFilter = startDate || endDate ? {
-    scheduledTime: {
-      ...(startDate && { gte: startDate }),
-      ...(endDate && { lte: endDate })
-    }
-  } : {}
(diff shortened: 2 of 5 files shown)


TusharW4ni:

| Gate | Result |
|---|---|
| Independent reviewer suite | ✅ 71/71 green (~33s) |
| Revert-check | ✅ end-day afternoon ride excluded without the fix (hours delta 0, topRiders empty), counted with it |
| Behavior parity (3 endpoints, different date logic) | ✅ topRiders YTD default byte-identical; completionRate/hours no-params still all-time; open bounds preserved |
| End-of-day + rollover + UTC | ✅ `lt: start-of-next-day`, month/year rollover correct, parse/filter both UTC — no skew |
| Scope / risk files | ✅ shared helper + 3 endpoints + test, +46 net; none |

Non-blocking notes are pre-existing (unvalidated date string; YTD default local-time) — not introduced here, left out of scope.

PR #59, opened 2026-07-08
Fix #9: format notification dates with Intl instead of comma-splitting

### LINKED ISSUE #9: [LOW] Inconsistent/Unsafe Native Date Conversions
### Location
`server/api/post/rides/index.ts`

### Description
Parsing the date using `formattedTime.split(',')[0]` after calling `toLocaleString()` will behave differently based on the server environment locale. With `dateStyle: 'full'`, the output includes the day of the week (e.g., "Monday, February 20..."). Splitting by a comma extracts "Monday" instead of an actual date string.

### Recommended Fix
Use standard UTC/ISO formatting strings, or a native formatting library (`Intl.DateTimeFormat`) instead of slicing commas.

TusharW4ni:
## What was broken

Notification emails built their `{{date}}` field with `toLocaleString(..., { dateStyle: 'full' })` followed by `formattedTime.split(',')[0]`. With `dateStyle: 'full'` the output looks like `"Monday, February 16, 2026 at 9:00 AM"`, so `split(',')[0]` extracts the **weekday** (`"Monday"`) — not a date — into `{{date}}`. The approach was also fragile to server locale/format changes. The same comma-splitting pattern was duplicated across three notification-building sites.

## What changed

- Added a pure, auto-imported helper `server/utils/datetime.ts` exposing `formatNotificationDate(date)` and `formatNotificationTime(date)`, built on `Intl.DateTimeFormat` with explicit `year/month/day` and `hour/minute` options and a pinned `timeZone: 'America/Chicago'` (matching the intent already present in `post/rides`).
- Swapped all three notification sites to use it: `server/api/post/rides/index.ts`, `server/api/post/rides/[id]/signup.ts`, `server/api/put/rides/[id].ts`. Now `{{date}}` is a real calendar date and `{{time}}` a real clock time regardless of host locale.

Scope is limited strictly to notification date/time formatting — no other notification logic, email send, schema, auth, CI, docker, or deps touched.

## Verification

**Email-only bug, tested via a pure helper.** The `date`/`time` values flow only into notification emails, which the e2e harness swallows — there is no API response to assert against. So the helper is pure and unit-tested directly in `tests/e2e/notification-datetime.test.ts` (imports the helper, no `setup()`/boot — same pattern as `trusted-origins.test.ts`, issue #21).

Tests in `tests/e2e/notification-datetime.test.ts`:
- `formats a real calendar date, not a bare weekday` — for `new Date('2026-02-16T15:00:00Z')` (a Monday), asserts the result matches `/February/` and `/16/` and is **not** `"Monday"`. This pins the bug.
- `formats a real clock time` — asserts the result matches `/\d{1,2}:\d{2}/` (and specifically `9:00 AM` for the Chicago-time conversion).

**Pre-fix failing observation / revert-check shape:** the helper does not exist before the fix, so the test file fails to import. Observed pre-fix (helper moved away):
```
FAIL  tests/e2e/notification-datetime.test.ts
Error: Cannot find module '../../server/utils/datetime' imported from .../tests/e2e/notification-datetime.test.ts
```
This is a **helper-absence** revert-check (like #21): reverting the fix removes the helper and the test stops collecting. Reviewers/gate should expect the import error, not an assertion diff, when the non-test change is reverted.

- `pnpm test`: 16 files, **73 tests passed**.
- `pnpm build`: succeeds.

Fixes #9

### DIFF
--- a/server/api/post/rides/[id]/signup.ts
+++ b/server/api/post/rides/[id]/signup.ts
@@ -118,11 +118,11 @@ export default defineEventHandler(async (event) => {
 
   // To Volunteer
   if (volunteerEmail) {
-    const formattedTime = new Date(ride.scheduledTime).toLocaleString()
+    const scheduledTime = new Date(ride.scheduledTime)
     notifications.push(sendNotification('RIDE_ASSIGNED', volunteer.id, {
       name: volunteer.user.name,
-      date: formattedTime.split(',')[0],
-      time: formattedTime,
+      date: formatNotificationDate(scheduledTime),
+      time: formatNotificationTime(scheduledTime),
       client: ride.client.user.name,
       pickup: ride.pickupDisplay,
       dropoff: ride.dropoffDisplay,
--- a/server/api/post/rides/index.ts
+++ b/server/api/post/rides/index.ts
@@ -52,17 +52,13 @@ export default defineEventHandler(async (event) => {
   })
 
   // Notify all volunteers
-  const formattedTime = new Date(body.scheduledTime).toLocaleString('en-US', {
-    timeZone: 'America/Chicago',
-    dateStyle: 'full',
-    timeStyle: 'short',
-  })
+  const scheduledTime = new Date(body.scheduledTime)
 
   await broadcastNotification('RIDE_CREATED', {
     pickup: pickupDisplay,
     dropoff: dropoffDisplay,
-    date: formattedTime.split(',')[0], // Approximate date
-    time: formattedTime,
+    date: formatNotificationDate(scheduledTime),
+    time: formatNotificationTime(scheduledTime),
     link: `${process.env.APP_URL || 'http://localhost:3000'}/rides/${ride.id}`,
   })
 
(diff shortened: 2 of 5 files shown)


TusharW4ni:

| Gate | Result |
|---|---|
| Independent reviewer suite | ✅ 73/73 green (~27s) |
| Revert-check | ⚠️ weak by nature (helper-absence import error — email-only bug, no API path) — **compensated below** |
| Usage completeness (the real check) | ✅ all 3 sites (post/rides:60-61, signup:119-120, put/rides:123-124) use the helper for date AND time; `split(',')`/`dateStyle:'full'` fully removed (grep-confirmed) |
| Old bug demonstrated | ✅ old expression yields "Monday"; helper yields "February 16, 2026" |
| Test quality | ✅ asserts real date (/February/, /16/, not weekday) + time pattern — catches a weekday-only regression |
| Scope / risk files | ✅ 3 endpoints + helper + test, +88/-13; none. Bonus: signup/put now TZ-consistent |

PR #60, opened 2026-07-09
Fix #10: require totalRideTime when completing a ride (server-side)

### LINKED ISSUE #10: [LOW] UI Mismatch on Total Ride Time Logic
### Location
`app/pages/rides/[id].vue` and `server/api/put/rides/[id].ts`

### Description
The UI asks the user to enter `totalRideTime` when marking a ride as "Completed", but the backend does not structurally validate or enforce that `totalRideTime` actually exists before triggering the 'Ride Completed by Volunteer' admin email, which results in "Total Time: N/A" being emailed.

### Recommended Fix
Require `totalRideTime` in the backend validation schema when the status is updated to `COMPLETED`.

TusharW4ni:
## What was broken

`PUT /api/put/rides/[id]` accepted `status: 'COMPLETED'` with no `totalRideTime`. The Complete modal in `app/pages/rides/[id].vue` already enforces `totalRideTime >= 0.1` via zod, so the normal UI flow is safe — but a direct API call (or any future UI path) could complete a ride with null hours, producing "Total Time: N/A" in the admin completion email and a completed ride with a null duration that skews the metrics from #18.

## What changed

In `server/api/put/rides/[id].ts`, after fetching `existingRide`, when the update sets `status: 'COMPLETED'` the endpoint now requires a valid `totalRideTime` (>= 0.1, matching the frontend rule) — taken from the request body if present, otherwise from the value already stored on the ride. If neither is a valid positive duration, it rejects with **400** and a clear message.

**Rule for already-completed rides:** re-updating a ride that is already COMPLETED (e.g. editing its notes) without re-sending `totalRideTime` is allowed, because the existing stored duration satisfies the requirement. Only a completion with no valid duration anywhere is rejected.

No changes to the volunteerId/#7 logic, notifications, or date formatting. No schema/auth/CI/docker/deps changes.

## Verification

New regression test: `tests/e2e/complete-requires-ridetime.test.ts`
- `rejects COMPLETED with no totalRideTime (400) and does not complete the ride` — asserts 400 and that the ride stays ASSIGNED.
- `rejects COMPLETED with a totalRideTime below 0.1 (400)`
- `accepts COMPLETED with a valid totalRideTime (200)` — the normal complete flow still succeeds.
- `allows re-updating an already-completed ride without re-sending totalRideTime` — documents the stored-value rule.

**Pre-fix failing assertion (observed before the fix):** `AssertionError: promise resolved "{ …(16) }" instead of rejecting` on `rejects.toMatchObject({ statusCode: 400 })` — i.e. completing with no duration returned 200 and marked the ride COMPLETED. Both rejection tests pass after the fix.

Also updated one assertion in `tests/e2e/unassign-status.test.ts` ("respects an explicit status when unassigning") to send `totalRideTime: 1.5` — it previously completed a ride with no duration, which the new server rule correctly forbids; the #7 behavior it tests is unchanged.

Suite: 77 passed (17 files), stable across two runs. `pnpm build` succeeds.

Fixes #10

### DIFF
--- a/server/api/put/rides/[id].ts
+++ b/server/api/put/rides/[id].ts
@@ -55,6 +55,24 @@ export default defineEventHandler(async (event) => {
     })
   }
 
+  // Issue #10: a ride can only be marked COMPLETED with a valid duration.
+  // The Complete modal (app/pages/rides/[id].vue) already enforces
+  // totalRideTime >= 0.1 via zod, but a direct API call could complete a ride
+  // with no duration — producing "Total Time: N/A" in the admin email and a
+  // completed ride with null hours that skews the metrics (#18). Require a
+  // valid totalRideTime either on this request OR already stored on the ride
+  // (so re-updating an already-completed ride without re-sending it still works).
+  if (body.status === 'COMPLETED') {
+    const effectiveRideTime =
+      body.totalRideTime !== undefined ? body.totalRideTime : existingRide.totalRideTime
+    if (effectiveRideTime == null || effectiveRideTime < 0.1) {
+      throw createError({
+        statusCode: 400,
+        statusMessage: 'totalRideTime (at least 0.1 hours) is required to complete a ride',
+      })
+    }
+  }
+
   // Build the Prisma update payload only from validated, whitelisted fields.
   const updateData: {
     status?: 'CREATED' | 'ASSIGNED' | 'COMPLETED'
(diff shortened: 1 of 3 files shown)


TusharW4ni:

| Gate | Result |
|---|---|
| Independent reviewer suite | ✅ 77/77 green (~38s) |
| Revert-check | ✅ 'no totalRideTime' + 'below 0.1' cases return 200 (+persist `totalRideTime:0`) without the fix, 400 with it |
| Cross-test edit | ✅ legitimate — only adds `totalRideTime:1.5` to the one #7 case that completed with no duration; no assertion weakened; no other COMPLETED-without-duration test missed |
| Bypass audit | ✅ string/null rejected by zod; 0/negative caught by `< 0.1`; fresh ride w/o duration → 400 |
| Scope / risk files | ✅ +18 source in one endpoint + tests; none |

Issue #62, opened 2026-07-09
E2E shared seeded DB is order-dependent — mutating tests can flake (harness robustness)


TusharW4ni:
### Surfaced during review of #11 (PR #61)

The #45 harness runs all `tests/e2e/*` files sequentially against ONE shared, seeded SQLite DB. Mutating tests (create/complete/delete rides, clear phones, etc.) must clean up after themselves, and several assert against seeded rows. This makes the suite **order-dependent**: an author saw a one-off failure in `complete-requires-ridetime.test.ts` attributed to shared-DB ordering (it did NOT reproduce across 3 subsequent full runs, so it's latent, not chronic).

Because the auto-merge gate requires a green full suite, a real intermittent flake here would block/mislead the loop.

### Options
- **Re-seed (or reset) the DB between test files** (globalSetup already builds the app once; add a per-file DB reset via a beforeAll that truncates+reseeds, or run seed per file against a per-file DB copy).
- **Wrap each test in a transaction that rolls back** (needs a Prisma/adapter hook; harder with the HTTP-server-in-separate-process model).
- **Make mutating tests fully self-contained** (create their own entities, never assert global counts against seed) — partially done already.
- At minimum, add a short retry for known-flaky suites and `log()` when it retries, so flakes are visible rather than silent.

Follow-up to #45. Low priority while the flake is not reproducing, but worth hardening before the suite grows much larger.

PR #63, opened 2026-07-09
Fix #22: drop the un-toggleable Cancelled filter and sanitize stale saved filters

### LINKED ISSUE #22: [UI/Logic] Un-toggleable 'Cancelled' Status Filter in Rides Dashboard
### Location
`app/pages/rides/index.vue` (Filter Dropdown UI)

### Description
In the Rides dashboard, the "Exclude Status" dropdown visually displays "Cancelled" in its active header summary, but the actual list of checkboxes inside the dropdown is missing the "Cancelled" option, making it impossible for users to (de)select it.

### Connection to Issue #5
This UI disconnect is a direct consequence of the backend schema issue reported in **Issue #5**. Because `CANCELLED` is missing from the `RideStatus` enum, it is omitted from the dynamically computed `filterOptions` array. However, `status:CANCELLED` is hardcoded into the default `savedExcludedFilters` cookie state, trapping the UI in an inconsistent state where the filter is active but un-manageable.

### Recommended Fix
Once Issue #5 is resolved (adding `CANCELLED` to the Prisma enum), ensure the UI `filterOptions` correctly loops over or includes the updated status.

TusharW4ni:
## What was broken
On the Rides dashboard (`app/pages/rides/index.vue`), the "Exclude Status" dropdown defaulted to excluding `status:CANCELLED`. But `CANCELLED` is not a real `RideStatus` (only `CREATED | ASSIGNED | COMPLETED` exist) and so is never in the toggleable `filterOptions`. The result: "Cancelled" showed as an active exclusion in the dropdown header with no checkbox to turn it off — a stuck, un-manageable filter. This is **frontend-only** and independent of #5 (whether CANCELLED ever becomes a real status); removing a filter for a nonexistent status is correct today.

## What changed
- New pure helper `app/utils/rideFilters.ts`:
  - `BASE_FILTER_OPTIONS` — the base status options (single source of truth).
  - `DEFAULT_EXCLUDED_FILTERS` — the corrected default, excluding **only** `Completed` (the sensible, toggleable default). The stale `Cancelled` entry is gone.
  - `sanitizeSavedFilters(saved, validValues)` — drops any saved filter whose value isn't among the known/toggleable values.
- `app/pages/rides/index.vue` now:
  - uses `DEFAULT_EXCLUDED_FILTERS` for the `ride-excluded-filters` cookie default (fixes new users), and
  - **sanitizes on load** — initializes `activeFilters`/`excludedFilters` via `sanitizeSavedFilters(...)`, so **existing users** whose `ride-excluded-filters` cookie already contains `status:CANCELLED` are freed, not just new users.
  - reuses `BASE_FILTER_OPTIONS` in `filterOptions` so the option list and the sanitize whitelist can't drift.
- The Created/Assigned/Completed include/exclude filtering, the `assign:ME` volunteer filter, and the "Completed" default exclusion are all preserved (`assign:ME` is included in the known-values whitelist).

## Verification
- New pure unit test `tests/e2e/ride-filters.test.ts` (imports the helper directly; does **not** boot the server, mirroring `trusted-origins.test.ts`):
  - `has no default excluded filter that is absent from the toggleable options` — pins the bug; **fails pre-fix** (default included the un-toggleable `status:CANCELLED`), passes after.
  - `keeps the sensible Completed default exclusion`
  - `strips the stale status:CANCELLED entry but keeps valid ones`
  - `keeps runtime-valid values that are passed in validValues (e.g. assign:ME)`
  - `handles null/undefined saved cookie gracefully`
- Confirmed the pinning test **fails** when the buggy `status:CANCELLED` default is reintroduced, and **passes** after the fix.
- `pnpm test` — 19 files / 84 tests green (including the new ones).
- `pnpm build` — succeeds.

No schema/auth/CI/docker/deps changes. No risk files touched.

Fixes #22

### DIFF
--- a/app/pages/rides/index.vue
+++ b/app/pages/rides/index.vue
@@ -3,6 +3,11 @@
   import type { TableColumn, TableRow } from '@nuxt/ui'
   import * as z from 'zod'
   import { authClient } from '../../utils/auth-client'
+  import {
+    BASE_FILTER_OPTIONS,
+    DEFAULT_EXCLUDED_FILTERS,
+    sanitizeSavedFilters,
+  } from '../../utils/rideFilters'
 
   const UBadge = resolveComponent('UBadge')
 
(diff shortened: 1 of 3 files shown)


TusharW4ni:

| Gate | Result |
|---|---|
| Independent reviewer suite | ✅ 84/84 green (~32s); no #62 flake this run |
| Revert-check | ✅ two variants: helper-absent fails; **reintroducing the buggy CANCELLED default also fails** the pinning assertion; both pass with the fix |
| Helper actually wired | ✅ corrected default (`DEFAULT_EXCLUDED_FILTERS`, no CANCELLED) AND on-load `sanitizeSavedFilters` for both active+excluded → frees existing stuck users |
| `assign:ME` preserved | ✅ `knownFilterValues` includes it, so a volunteer's saved filter isn't over-stripped |
| Scope / risk files | ✅ 3 files, +65/-12; filter-application logic untouched; none |

PR #64, opened 2026-07-09
Fix #53: store blank admin phone numbers as null

### LINKED ISSUE #53: Admin create/update endpoints have the same blank-phone unique-constraint bug (#15 sibling)
### Discovered during review of #15 (PR #52)

#15 sanitized blank phone → null on the client/volunteer create/update endpoints, but the **admin** endpoints still pass `phone: body.phone` unsanitized:
- `server/api/post/admins/index.ts` (create + update branches, ~lines 25 & 34)
- `server/api/put/admins/[id].ts` (~line 19)

So clearing a second admin's phone (sending `""`) will still trip the `User.phone @unique` constraint → P2002/500 — the identical bug #15 fixed elsewhere.

### Fix
Apply the existing `emptyToNull()` helper (`server/utils/sanitize.ts`, added in #52) to the phone writes in both admin endpoints. Trivial, mirrors #15.

TusharW4ni:
## What was broken

The admin create/update endpoints (`server/api/post/admins/index.ts`, `server/api/put/admins/[id].ts`) still wrote `phone: body.phone` unsanitized — the sibling of #15. Clearing a second admin's phone (the frontend sends `""`) stores a duplicate blank string, which trips the `User.phone @unique` constraint (SQLite treats `""` as a distinct value), returning P2002 → 500. #15 fixed this for the client/volunteer endpoints but not admins.

## What changed

Applied the existing `emptyToNull()` helper (`server/utils/sanitize.ts`, added in #52) to every phone write in the two admin endpoints — the create branch, the existing-user upgrade branch, and the PUT update. Blank/whitespace phones are now stored as `null`, so any number of admins can have "no phone". No schema/auth/deps/CI changes.

## Verification

New regression test `tests/e2e/admin-empty-phone.test.ts`:
- Creates two throwaway admins (unique emails), clears both their phones via the admin PUT endpoint; the second clear 500s pre-fix (P2002), both succeed post-fix and store `null` (asserted `null`, not `""`).
- Creates an admin with a blank phone via POST; 500s pre-fix, stores `null` post-fix.
- Cleans up via `/api/delete/admins/<id>` so the shared seeded DB is undisturbed.

Confirmed FAIL pre-fix (both cases 500), PASS post-fix. Full suite green (`pnpm test` → 20 files, 86 tests) and `pnpm build` succeeds.

Fixes #53

### DIFF
--- a/server/api/post/admins/index.ts
+++ b/server/api/post/admins/index.ts
@@ -22,7 +22,7 @@ export default defineEventHandler(async (event) => {
       data: {
         role: 'ADMIN',
         name: body.name,
-        phone: body.phone,
+        phone: emptyToNull(body.phone),
       },
     })
   } else {
@@ -31,7 +31,7 @@ export default defineEventHandler(async (event) => {
       data: {
         name: body.name,
         email: body.email,
-        phone: body.phone,
+        phone: emptyToNull(body.phone),
         role: 'ADMIN',
       },
     })
--- a/server/api/put/admins/[id].ts
+++ b/server/api/put/admins/[id].ts
@@ -16,7 +16,7 @@ export default defineEventHandler(async (event) => {
     data: {
       name: body.name,
       email: body.email,
-      phone: body.phone,
+      phone: emptyToNull(body.phone),
     },
   })
 })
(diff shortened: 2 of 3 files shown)


TusharW4ni:

| Gate | Result |
|---|---|
| Independent reviewer suite | ✅ 86/86 green (~33s); no #62 flake |
| Revert-check | ✅ PUT second-clear → 500 (P2002) and POST-create → stores `""` without the fix; both null with it |
| Coverage | ✅ all 3 phone-write sites sanitized (post/admins create + upgrade branches, put/admins) |
| Scope / risk files | ✅ +3/-3 net non-test; none |

Closes out milestone 2 (Data Integrity).

Issue #24, opened 2026-04-02
[MEDIUM] N+1 Query Problem in Top Riders Endpoint


TusharW4ni:
### Location
`server/api/get/metrics/topRiders/index.ts`

### Description
The API uses a `Promise.all` loop executing `prisma.client.findUnique` sequentially/in parallel for every single grouped rider returned by the aggregate query. While this is currently capped at 5 (`take: 5`), it is an architectural anti-pattern that places unnecessary load on the database connection pool.

### Recommended Fix
Extract the resulting `clientId` array from the grouped query and perform a single batched `prisma.client.findMany({ where: { id: { in: clientIds } } })` query to resolve all relationships in one round-trip.

Issue #13, opened 2026-04-02
[HIGH] Unbounded Database Queries Without Pagination


TusharW4ni:
### Location
`server/api/get/rides/index.ts`

### Description
The endpoint calls `prisma.ride.findMany()` with nested joins (`client.user`, `volunteer.user`) but absolutely no pagination (`take`/`skip`). Because the frontend (`app/pages/rides/index.vue`) handles date filtering purely on the client side, the API will fetch the **entire history of all rides** into memory every time the dashboard loads. This will severely degrade server performance and freeze the frontend as the app scales.

### Recommended Fix
Move date range and status filtering to the backend Prisma query, and implement server-side pagination.

TusharW4ni:
🤖 **Flagged by `/issue-loop` — needs human-guided handling, not autonomous auto-deploy.**

This isn't blocked, but it's a poor fit for the auto-merge gate:
- The real fix couples a **backend change** (get/rides accepts pagination + server-side date/status filtering) with a **frontend rewrite** of the core rides dashboard (`app/pages/rides/index.vue` currently filters client-side in a computed). You can't half-do it — once results are paginated, client-side filtering would only filter the current page, so filtering MUST move server-side too.
- That frontend data-flow change auto-deploys to production on merge, and the e2e harness can't verify Vue component filtering/pagination behavior (it tests API + SSR, not client reactivity) — so an auto-merge would ship the main dashboard's largest change with only API-level verification.

**Recommendation:** do this as a deliberate, human-reviewed PR. I can author it in a focused session (backend pagination/filter params + adapting the dashboard, with browser verification), but it shouldn't ride the autonomous gate. High value (P1) — worth scheduling.

PR #65, opened 2026-07-09
Add query-count test seam to pin N+1 / unbounded-query regressions (#45)

### LINKED ISSUE #45: Test infrastructure: per-file rebuild/boot is slow, and no way to test concurrency (blocks #47)

TusharW4ni:
Extends the test harness so **performance** issues become gate-verifiable — the capability gap that was forcing every Milestone-3 issue to flag.

## Why
N+1 fixes (#24) and unbounded-query fixes (#13) change **no observable output** — topRiders returns identical data whether it does N+1 queries or one batched query. So a regression test passes with or without the fix and the loop's revert-check is inconclusive. This is the same limitation that blocked #12 until the race seam; this adds the analogous seam for query counts.

## What
A **test-only** per-request Prisma query counter (all strict no-ops in production):
- `server/utils/prisma.ts` — when `TEST_HOOKS=1`, a `$extends` query hook increments `queryCounter.count` on every operation (incl. inside transactions). Without the flag, the **base client is exported unchanged**.
- `server/middleware/zz-test-query-count.ts` — resets the counter per request when `x-test-count-queries` is sent. Named `zz-` so it runs **after** `auth.ts`, excluding the session-lookup queries from the count.
- `server/plugins/test-query-count.ts` — returns the count via the `x-query-count` response header (opt-in only).
- `tests/utils/harness.ts` — `fetchWithQueryCount()` helper returning `{ status, queryCount, body }`.
- Unifies the test-hooks flag: `TEST_RACE_HOOKS` → `TEST_HOOKS` (raceDelay + global-setup), so the race seam and query seam share one gate.

## Verification (`tests/e2e/query-count-seam.test.ts`)
- `get/addresses?search` issues **exactly 1** query (precise counting; auth excluded).
- `get/rides` stays **≤ 5** despite nested client/volunteer includes over many rows (bounded, not per-row N+1).
- `get/metrics/topRiders` is detected as **> 2** queries — proving the seam catches the real #24 N+1. (When #24 lands, that assertion tightens to `≤ 2`.)
- Header is **absent** without the opt-in.
- Full suite: **90/90 green**; `pnpm build` succeeds; race seam unaffected under the renamed flag.

## Production safety
Without `TEST_HOOKS` (i.e. always, outside the harness): the base Prisma client is used, the middleware and plugin return immediately, and a hostile client sending `x-test-count-queries`/`x-query-count` gets nothing. The only prod-file change is `prisma.ts` (guarded ternary) — behavior identical when the flag is unset.

## Unlocks
#24 (assert topRiders ≤ 2), and the query-efficiency aspect of #13, plus any future N+1 — now with a genuine failing-pre-fix test that clears the revert-check.

Fixes #45

### DIFF
--- a/server/middleware/zz-test-query-count.ts
+++ b/server/middleware/zz-test-query-count.ts
@@ -1,0 +1,13 @@
+import { queryCounter } from '../utils/prisma'
+
+// Query-count seam (issue #45). Resets the per-request query counter so an e2e
+// test can measure only the handler's queries. Named `zz-` so it runs AFTER
+// `auth.ts` (Nitro runs middleware in alphabetical order) — the auth session
+// lookup's queries happen first and are excluded from the count. No-op unless
+// TEST_HOOKS=1 and the request opts in with `x-test-count-queries`.
+export default defineEventHandler((event) => {
+  if (process.env.TEST_HOOKS !== '1') return
+  if (getHeader(event, 'x-test-count-queries')) {
+    queryCounter.count = 0
+  }
+})
--- a/server/plugins/test-query-count.ts
+++ b/server/plugins/test-query-count.ts
@@ -1,0 +1,15 @@
+import { queryCounter } from '../utils/prisma'
+
+// Query-count seam (issue #45). After the handler runs, expose the number of
+// Prisma queries it issued via the `x-query-count` response header so e2e tests
+// can assert bounds (e.g. topRiders must issue ≤ 2 queries, not N+1). Paired
+// with the zz-test-query-count middleware that resets the counter per request.
+// No-op in production (TEST_HOOKS unset).
+export default defineNitroPlugin((nitro) => {
+  if (process.env.TEST_HOOKS !== '1') return
+  nitro.hooks.hook('beforeResponse', (event) => {
+    if (getHeader(event, 'x-test-count-queries')) {
+      setResponseHeader(event, 'x-query-count', String(queryCounter.count))
+    }
+  })
+})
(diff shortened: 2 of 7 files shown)


TusharW4ni:
🤖 **Independent review: RECOMMEND_MERGE** (verified in a clean worktree by a reviewer with no author context).

| Check | Result |
|---|---|
| Full suite | ✅ 21 files / 90 tests, ~31s; new seam test green; **signup-race still green** (rename didn't break the race seam) |
| Counting accuracy + middleware order | ✅ `zz-` sorts after `auth.ts`, so auth's session queries are excluded — `addresses === 1` is accurate & non-flaky (the `zz-` naming is load-bearing) |
| Prod no-op (critical) | ✅ flag unset → base client exported unchanged; middleware/plugin early-return; prod client sending the header gets nothing |
| `$extends` / transaction safety | ✅ pure pass-through (`count++; return query(args)`) — transactions + better-auth adapter behave identically |
| Catches N+1 | ✅ topRiders (groupBy + findUnique/client) detected as >2; batched #24 fix drops it to ≤2 |
| Global-counter caveat | ✅ all calls sequentially awaited; no `Promise.all` of counted requests |
| Scope / risk | ✅ +146/-18; only prod file is `prisma.ts` (flag-guarded); no schema/CI/docker/deps/auth-config |

Held for human merge (touches `prisma.ts`; merge = prod deploy). Recommended — unlocks #24 and the N+1/unbounded-query class for the loop.

Issue #32, opened 2026-07-04
Ride Creation Blocks on Sequential Email Broadcast to All Volunteers


TusharW4ni:
### Location
`server/api/post/rides/index.ts` (line ~68) and `server/utils/notification.ts` (`broadcastNotification`)

### Description
Creating a ride does `await broadcastNotification('RIDE_CREATED', ...)` before responding. `broadcastNotification` loops over **every AVAILABLE volunteer sequentially**, and each iteration runs 2 DB queries plus a full synchronous SMTP send through Gmail (`sendNotification` → `sendEmail` → `transporter.sendMail`).

With N available volunteers, the admin who clicks "Create Ride" waits for N sequential SMTP round-trips (~1s each) before the modal closes. At 30 volunteers that's ~30 seconds of hang, likely a frontend timeout, and — because there's no idempotency — retries send everyone duplicate emails. Also note `sendEmail` swallows failures (#25), so partial delivery is invisible.

### Recommended Fix
Respond immediately after the DB write and dispatch emails asynchronously:
- Minimal: don't `await` the broadcast (fire-and-forget with `.catch` logging), and fetch the template once instead of per-volunteer.
- Better: batch with `Promise.allSettled` over the volunteer list (as `signup.ts` already does for its notifications), or move broadcasts into the existing cron/queue machinery from #17.

The same sequential-loop pattern is why the reminder cron blocks (#17) — a shared fix is ideal.

TusharW4ni:
🤖 **Skipped by `/issue-loop` — needs the same fault-injection seam as #17.**

The fix (make `post/rides` fire-and-forget the volunteer broadcast instead of `await`ing N sequential sends) changes **no observable output** — the response returns the created ride either way. The only differences are timing (response latency) and error-propagation (a broadcast failure currently blocks/fails the create; after the fix it shouldn't). So it can't be pinned by an output test, and the revert-check is inconclusive — same class as #17.

Two ways to make it gate-verifiable, neither available yet:
1. A **fault-injection seam** (throw inside the broadcast when a test header is set): pre-fix the `await` propagates → 500; post-fix fire-and-forget → 200. Deterministic. This is the SAME seam #17 needs.
2. A **timing test** using the existing race-delay seam (inject per-send delay, assert the response returns before N×delay) — but timing assertions are flaky in CI, so I won't ship one.

[...]

Issue #17, opened 2026-04-02
[MEDIUM] Cron Job Blocking / Notification Race Condition


TusharW4ni:
### Location
`server/utils/scheduler.ts` and `server/cron/reminders.ts`

### Description
The reminder cron job processes emails synchronously (`await sendNotification` inside a loop). If SMTP is slow or there are many reminders, the cron tick blocks and piles up. Furthermore, `prisma.sentReminder.create` occurs *after* the email sends. If the server crashes or Prisma errors out right after the email sends, the system loses state and will re-send the same email indefinitely every 5 minutes.

### Recommended Fix
Use a transaction to write the `SentReminder` record *before* or concurrently with dispatching the email, or move to a background queue system.

TusharW4ni:
🤖 **Skipped by `/issue-loop` — needs a harness capability not yet built.**

#17 has two parts. The **blocking** part (sequential `await` over reminders) overlaps #32's fire-and-forget approach. The **race/crash-recovery** part is the hard one: the fix (write `SentReminder` before/transactionally with the send) only changes behavior when the process **crashes between the email send and the DB write** — the happy path is already idempotent (second cron tick sees the SentReminder and skips), so a normal test passes with or without the fix, and the revert-check is inconclusive.

Pinning it needs two test-only seams the harness doesn't have yet:
1. an **on-demand trigger** for `processReminders` (it runs on a cron timer, not an HTTP request, so a test can't invoke it), and
2. **fault injection** to simulate a crash after send / before `SentReminder.create`.

This is the same class of gap that the query-count seam (#65) just filled for N+1. I can build a fault-injection + reminder-trigger seam (like #45/#65) if you want #17 auto-mergeable — say the word. Until then it should be human-guided.

TusharW4ni:
🤖 Un-flagged: the fault-injection seam (#67) is merged, so this is now cleanly testable (arm a fault between the email send and the SentReminder write; assert no re-send after the fix). Re-entering the `/issue-loop` queue.

PR #67, opened 2026-07-09
Add fault-injection test seam to unlock #17 and #32 (#45)

### LINKED ISSUE #45: Test infrastructure: per-file rebuild/boot is slow, and no way to test concurrency (blocks #47)

TusharW4ni:
Third harness capability (after the race seam #12 and query-count seam #65) — this one unlocks the two remaining Milestone-3 issues whose fixes change **no observable output**.

## Why
- **#17** (cron notification race): the fix (write `SentReminder` before/with the send) only matters if the process crashes between send and DB-write — the happy path is already idempotent, so an output test passes either way.
- **#32** (broadcast blocking): fire-and-forget vs `await` returns the same response; only latency/error-propagation differ.

Neither can clear the revert-check without a way to inject a mid-operation failure. This adds that.

## What
- `server/utils/testHooks.ts` — `maybeFault(label, event?)` with two triggers, both **no-ops in production** (require `TEST_HOOKS=1`):
  - **request-scoped**: fires when `x-test-fault: <label>` matches (for request-reachable code, e.g. #32's broadcast).
  - **context-free**: fires when `armFault(label)` was called (for code with no request event, e.g. the cron scheduler #17). `armFault`/`disarmFault` bracket a run.
- `server/api/_test/run-reminders.post.ts` — **hard-gated (404 unless `TEST_HOOKS=1`)** endpoint that invokes `processReminders()` on demand (the job otherwise only runs on a cron timer, unreachable from a test) and can arm a fault via `{ fault }` body.
- `server/utils/scheduler.ts` — one `maybeFault('reminders-scan')` hook (no-op in prod); #17 can add a finer point between send and the SentReminder write.

## Verification (`tests/e2e/fault-injection-seam.test.ts`, 5 cases)
- on-demand reminder run with no fault → 200
- request-header fault (`x-test-fault: run-reminders`) → 500
- non-matching label → 200
- armed cron fault (`{ fault: 'reminders-scan' }`) fires inside `processReminders` → 500 (proves the context-free path)
- fault disarmed after the run → next run 200
- Full suite: **97/97**; `pnpm build` clean.

## Production safety
Without `TEST_HOOKS` (always, outside the harness): the `_test/run-reminders` endpoint 404s, `maybeFault` returns immediately, and `armFault` is never reachable. Prod files touched are `scheduler.ts` (one guarded no-op line) and `testHooks.ts` (test-only) — no behavior change when the flag is unset.

## Unlocks
#17 (arm a fault between send and SentReminder-write; assert no re-send after the fix) and #32 (inject a broadcast fault; pre-fix `await` → 500, post-fix fire-and-forget → 200). Both become clean, revert-checkable auto-merges.

Fixes #45 (final harness capability)

### DIFF
--- a/server/api/_test/run-reminders.post.ts
+++ b/server/api/_test/run-reminders.post.ts
@@ -1,0 +1,31 @@
+import { processReminders } from '../../utils/scheduler'
+import { armFault, disarmFault, maybeFault } from '../../utils/testHooks'
+
+// Test-only endpoint (issue #45). The reminder job runs on a cron timer, so an
+// e2e test can't otherwise invoke it. This lets a test drive processReminders()
+// on demand and optionally arm a fault (e.g. simulate a crash between the email
+// send and the SentReminder write) to pin the crash-recovery fix (#17).
+//
+// Hard-gated: returns 404 unless TEST_HOOKS=1 (only the harness sets it), so it
+// does not exist in production. Still behind the global auth middleware (admin).
+//
+// Body: { fault?: string } — arms that fault label for the duration of the run.
+export default defineEventHandler(async (event) => {
+  if (process.env.TEST_HOOKS !== '1') {
+    throw createError({ statusCode: 404, statusMessage: 'Not found' })
+  }
+
+  // Validates the request-scoped fault path (x-test-fault: run-reminders).
+  maybeFault('run-reminders', event)
+
+  const body = await readBody(event).catch(() => ({}))
+  const fault = typeof body?.fault === 'string' ? body.fault : undefined
+
+  if (fault) armFault(fault)
+  try {
+    await processReminders()
+    return { ok: true }
+  } finally {
+    if (fault) disarmFault(fault)
+  }
+})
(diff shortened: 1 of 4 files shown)


TusharW4ni:
🤖 **Independent review: RECOMMEND_MERGE** (verified in a clean worktree, no author context).

| Check | Result |
|---|---|
| Full suite | ✅ 23 files / 97 tests; new fault seam 5/5; race + query seams (share testHooks.ts) not regressed |
| Prod gating | ✅ endpoint 404s before any action when TEST_HOOKS unset; `maybeFault` returns immediately; `armedFaults` only ever armed by the gated endpoint (empty in prod); admin-gated by the global middleware |
| Seam correctness | ✅ header path exact-match 500/200; armed path fires inside processReminders and **propagates** (placed before the swallowing try/catch); `finally` disarms — no cross-request leak (test 5 proves it) |
| Scheduler safety | ✅ one guarded line; real cron unaffected with flag unset |
| Scope / risk | ✅ 4 files; only prod touches are the gated `_test` endpoint + one `scheduler.ts` no-op line; no schema/CI/docker/deps/auth-config |

Held for human merge (touches `scheduler.ts`; merge = prod deploy). Recommended — unlocks #17 and #32 as clean auto-merges.

PR #68, opened 2026-07-09
Fix #17: claim reminders before sending to prevent re-send on crash

### LINKED ISSUE #17: [MEDIUM] Cron Job Blocking / Notification Race Condition

TusharW4ni:
## What was broken

`processReminders` (`server/utils/scheduler.ts`) sent the reminder email with `sendNotification(...)` and only **then** wrote the `SentReminder` row. If the process crashed or Prisma errored between the send and the write, no row was recorded — so every subsequent cron tick (every 5 minutes) re-sent the same reminder indefinitely. This is the crash-recovery race called out in #17 (the P-driver; the secondary blocking concern is tracked separately).

## What changed

Reordered to **"claim before send"**:

1. `prisma.sentReminder.create({ data: { rideId, type } })` runs **first**, relying on the existing `@@unique([rideId, type])` to make the claim atomic/idempotent.
2. A `P2002` unique violation means a prior tick already claimed it → `continue` (skip, do **not** send). Detected with `instanceof Prisma.PrismaClientKnownRequestError && err.code === 'P2002'`, matching the pattern used elsewhere in `server/api/**`.
3. The claim is **outside** the send `try/catch`, so a real claim failure surfaces instead of being swallowed; the send keeps its own swallowing catch (per #25).

Result: **at-most-once** delivery. A crash after the claim leaves the row behind, so the next tick sees it and never re-sends. Missing a single email on a crash is acceptable per the issue — far better than infinite spam.

A test-only `maybeFault('reminder-after-claim')` seam (built on the #67 fault-injection seam) is placed between the claim and the send. It is a complete no-op in production (gated on `TEST_HOOKS=1`); it lets the e2e test simulate "claimed but crashed before the send finished."

No schema change (the unique constraint already existed). No auth/CI/docker/deps changes.

## Verification

**Regression test:** `tests/e2e/reminder-claim-before-send.test.ts` → `reminder claim-before-send (#17) > a faulted run leaves the reminder claimed and a later run does not re-send`.

The test creates a dedicated volunteer + `Reminder` (5-day window) + ASSIGNED future ride (unique IDs, cleaned up in `afterAll`) so a reminder is genuinely due, then:
1. Arms `reminder-after-claim` via `POST /api/_test/run-reminders { fault }` → simulates a crash mid-run.
2. Reads the `sent_reminder` table directly (better-sqlite3, same pattern as `tests/utils/auth.ts`) and asserts exactly **one** row exists after the faulted run.
3. Runs again with no fault and asserts the row count is **still one** — no re-send.

**Pre-fix failing assertion (observed before the fix):**
```
AssertionError: expected 200 to be 500
tests/e2e/reminder-claim-before-send.test.ts:150
```
Pre-fix, the injected fault (a simulated crash after the send, before the create) was swallowed by the send `try/catch`, so the run returned 200 and — critically — the `sent_reminder` row was never written (0 rows after the faulted run). Post-fix, the claim commits before the fault, so the faulted run leaves exactly one row and the endpoint surfaces the crash as 500. That row-lifecycle difference (row exists after a faulted run iff the fix is present) is the revert-check.

**Suite:** 98 passed (24 files), including the new test. `pnpm build` succeeds.

Fixes #17

### DIFF
--- a/server/utils/scheduler.ts
+++ b/server/utils/scheduler.ts
@@ -1,3 +1,4 @@
+import { Prisma } from '../../prisma/generated/client'
 import { prisma } from './prisma'
 import { sendNotification } from './notification'
 import { maybeFault } from './testHooks'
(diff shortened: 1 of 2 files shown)


TusharW4ni:

| Gate | Result |
|---|---|
| Independent reviewer suite | ✅ 98/98 green (~32s) |
| **Revert-check (via #67 fault seam)** | ✅ pre-fix: faulted run swallowed (200) + 0 sent_reminder rows → would re-send; post-fix: claim persists → 1 row, no re-send |
| Correctness | ✅ claim before send; P2002→skip; other errors rethrown; claim outside the send try/catch; at-most-once tradeoff is intended (send failure logged, not retried) |
| Scope / risk files | ✅ scheduler.ts + test, +30 net; no schema change (unique constraint pre-existed) |

Second perf/reliability fix pinned by a purpose-built seam — the fault seam (#67) working as intended.

PR #69, opened 2026-07-09
Fix #32: dispatch ride-created broadcast without blocking the response

### LINKED ISSUE #32: Ride Creation Blocks on Sequential Email Broadcast to All Volunteers

TusharW4ni:
## What was broken

`POST /api/post/rides` `await`ed `broadcastNotification('RIDE_CREATED', ...)` before returning, so the HTTP response blocked until every AVAILABLE volunteer had been emailed via N sequential synchronous SMTP round-trips. With N volunteers, the admin's "Create Ride" waited on N sends, and a single broadcast failure failed/blocked the whole create even though the ride row had already been written.

## What changed

- `server/api/post/rides/index.ts`: dispatch the broadcast **fire-and-forget** — the ride is created and returned to the client immediately, and the un-awaited broadcast's errors are swallowed with `.catch(console.error)` (mirrors the existing `Promise.allSettled(...).catch(...)` pattern in `signup.ts`). Response shape unchanged.
- `server/utils/notification.ts`: `broadcastNotification` now takes an optional `event?: H3Event` and calls `maybeFault('ride-broadcast', event)` at the start (the fault-injection seam from #67; no-op in production). Also parallelizes the per-volunteer sends with `Promise.allSettled` instead of a sequential `await` loop, so one failed send no longer aborts the rest.

No schema/auth/CI/docker/deps changes. `event` is optional so other `broadcastNotification` callers are unaffected.

## Verification

Regression test: `tests/e2e/ride-broadcast-nonblocking.test.ts`
- `a broadcast fault does not fail ride creation (200 + ride created)` — sends `POST /api/post/rides` with header `x-test-fault: ride-broadcast`.
- `a normal create (no fault) still returns 200 and creates the ride` — no-regression check.

**Revert-check (pinned via the #67 fault seam):**
- Pre-fix (awaited broadcast): the injected fault propagates out of the awaited call → create returns **500**. Observed failing assertion:
  `AssertionError: expected 500 to be 200` at `expect(status).toBe(200)`.
- Post-fix (fire-and-forget): the fault throws in the un-awaited broadcast, is caught by `.catch(console.error)`, and the request already returned **200** with the created ride.

Full suite green: **25 files / 100 tests passed**. `pnpm build` succeeds.

Fixes #32

### DIFF
--- a/server/api/post/rides/index.ts
+++ b/server/api/post/rides/index.ts
@@ -54,13 +54,17 @@ export default defineEventHandler(async (event) => {
   // Notify all volunteers
   const scheduledTime = new Date(body.scheduledTime)
 
-  await broadcastNotification('RIDE_CREATED', {
+  // Fire-and-forget the broadcast (issue #32): the ride is created, so return it
+  // immediately instead of blocking the response on N sequential SMTP sends. A
+  // slow or failing broadcast must not delay or fail ride creation, so we don't
+  // await it and swallow any error in the background (mirrors signup.ts).
+  broadcastNotification('RIDE_CREATED', {
     pickup: pickupDisplay,
     dropoff: dropoffDisplay,
     date: formatNotificationDate(scheduledTime),
     time: formatNotificationTime(scheduledTime),
     link: `${process.env.APP_URL || 'http://localhost:3000'}/rides/${ride.id}`,
-  })
+  }, event).catch(console.error)
 
   return ride
 })
--- a/server/utils/notification.ts
+++ b/server/utils/notification.ts
@@ -1,3 +1,4 @@
+import type { H3Event } from 'h3'
 import { prisma } from './prisma'
 import { sendEmail } from './email'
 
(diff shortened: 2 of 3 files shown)


TusharW4ni:

| Gate | Result |
|---|---|
| Independent reviewer suite | ✅ 100/100 green (~28s) |
| Revert-check | ✅ reproduced true pre-fix (seam wired + broadcast **awaited**) → fault propagates → 500; fire-and-forget → 200 (the literal-revert nuance handled correctly) |
| Unhandled-rejection safety | ✅ `.catch(console.error)` handles the async throw; ride created independent of broadcast |
| Scope / risk files | ✅ post/rides + notification.ts, +13/-4; `event` optional (only caller); none |

Third fault/perf fix pinned by a purpose-built seam. Milestone-3 backend is now essentially clear.

PR #70, opened 2026-07-09
Fix #19: debounced server-side address autocomplete

### LINKED ISSUE #19: [LOW] Autocomplete Feature Broken for Large Datasets
### Location
`app/pages/rides/index.vue` and `server/api/get/addresses/index.ts`

### Description
The address autocomplete frontend fetches a hardcoded `take: 20` addresses *once* on page mount, completely ignoring the user's search string. If the database has 100 addresses, it is impossible for users to search for or autocomplete any address that falls outside the first 20 alphabetically.

### Recommended Fix
Debounce the UI input and make dynamic requests passing `?search=xyz` to the server so the API can filter matches correctly.

TusharW4ni:
## What was broken
The create-ride modal's address autocomplete fetched `/api/get/addresses` **once on mount with no search param** and filtered that fixed list client-side. The backend caps results at `take: 20`, so any address outside the first 20 alphabetically was permanently unreachable via the pickup/dropoff search boxes (issue #19). The backend already supported `?search=` (`server/api/get/addresses/index.ts`); only the frontend wiring was missing.

## What changed
- `app/pages/rides/index.vue`: replaced the once-fetched-list + client-side filter with **debounced server-side queries**. Debounced watchers on `pickupSearch`/`dropoffSearch` call `$fetch('/api/get/addresses', { query: { search } })`, so suggestions come from the full table. Existing `onPickupSelect`/`onDropoffSelect`, the client `homeAddress` prefill, the dropdown UI, and form reset are unchanged.
- `app/utils/addressSearch.ts` (new): pure `buildAddressQuery(search)` (trims, returns `null` for empty terms so no fetch fires) and a `debounce()` helper — kept pure so the wiring is unit-testable (mirrors the existing `app/utils/rideFilters.ts` pattern).

No backend, schema, auth, CI, docker, or dependency changes.

## Verification

**Regression test** — `tests/e2e/address-search.test.ts` (pure unit test, no `setup()`, mirrors `trusted-origins.test.ts` / `ride-filters.test.ts`):
- `buildAddressQuery (#19)` — asserts the trimmed search term is carried into the request query and that empty/whitespace terms return `null` (no fetch).
- `debounce (#19)` — asserts single trailing invocation with the latest args, and `cancel()`.

**Pre-fix failing assertion (observed):** temporarily reverting the helper to the pre-fix behavior (never send a search param → `buildAddressQuery` returns `null`) makes the suite fail with:
```
AssertionError: expected null to deeply equal { search: 'Main St' }
  tests/e2e/address-search.test.ts > buildAddressQuery (#19) > carries the trimmed search term to the server query
```
The helper was then restored and the test passes.

**Honest caveat on the revert-check:** this is a **frontend reactivity** fix and the e2e harness is API/SSR-only — it cannot render or drive the Vue component's debounced reactive fetch. The regression test therefore pins the extracted pure logic (the query-building seam), **not** the component's `watch`→`$fetch` wiring directly. A revert of *only* the `.vue` `watch` wiring (leaving the helper intact) would not be caught by this test. Reviewer should treat the component wiring as human-reviewable.

**Live check of the server-side query path** (the query the component now calls): ran the endpoint's exact Prisma `findMany({ where: OR contains, take: 20 })` against a seeded dev DB — `search=Plano` returned only the 3 Plano rows; non-matching terms returned `[]`. Attempted a full browser drive but the dev server hit an `EMFILE: too many open files` file-watcher error in this worktree and the built (`.output`) server did not resolve the dev DB, so an end-to-end UI drive was not achievable here.

**Suite + build:** `pnpm test` → 26 files / 105 tests pass (incl. the new test). `pnpm build` succeeds.

Fixes #19

### DIFF
--- a/app/pages/rides/index.vue
+++ b/app/pages/rides/index.vue
@@ -108,21 +108,61 @@
   const state = reactive(blankRideForm())
 
   // --- Autocomplete Logic ---
+  // Query the backend with `?search=` (debounced) instead of fetching a fixed
+  // list of 20 once on mount and filtering it client-side. This lets any address
+  // in the table surface, not just the first 20 alphabetically (issue #19). The
+  // query-building and debounce logic live in a pure, unit-tested helper.
   const pickupSearch = ref('')
   const dropoffSearch = ref('')
 
-  const { data: addresses } = await useFetch('/api/get/addresses')
+  const pickupOptions = ref<any[]>([])
+  const dropoffOptions = ref<any[]>([])
 
-  const pickupOptions = computed(() => {
-    if (!pickupSearch.value || !addresses.value) return []
-    const q = pickupSearch.value.toLowerCase()
-    return addresses.value.filter((a: any) => a.label.toLowerCase().includes(q)).slice(0, 5)
+  async function fetchAddressOptions(search: string, target: Ref<any[]>) {
+    const query = buildAddressQuery(search)
+    if (!query) {
+      target.value = []
+      return
+    }
+    try {
+      const results = await $fetch<any[]>('/api/get/addresses', { query })
+      // Ignore stale responses: only apply if the input still matches the term
+      // we searched for (avoids out-of-order results overwriting newer ones).
+      if ((search ?? '').trim() === query.search) {
+        target.value = (results ?? []).slice(0, 5)
+      }
+    } catch (err) {
+      console.error('Failed to fetch address suggestions', err)
+      target.value = []
+    }
+  }
+
+  const debouncedPickupFetch = debounce(
+    (search: string) => fetchAddressOptions(search, pickupOptions),
+    250
+  )
+  const debouncedDropoffFetch = debounce(
+    (search: string) => fetchAddressOptions(search, dropoffOptions),
+    250
+  )
+
+  watch(pickupSearch, (term) => {
+    if (!buildAddressQuery(term)) {
+      // Clear immediately (and cancel any pending fetch) when the box empties.
+      debouncedPickupFetch.cancel()
+      pickupOptions.value = []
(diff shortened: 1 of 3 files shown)


TusharW4ni:
🤖 **Auto-review gate: FLAGGED — fix verified correct by inspection, but the component wiring can't be revert-checked. One-click merge recommended.**

This is the frontend-reactivity ceiling (the API/SSR harness can't drive a Vue `watch`→`$fetch`), not a defect. Independent reviewer verified everything that's verifiable:

| Check | Result |
|---|---|
| Full suite | ✅ 105/105 green, no regressions |
| Helper test genuine | ✅ reverting `buildAddressQuery` to no-op → test FAILS; restored → passes (not vacuous) |
| Component wiring (inspection) | ✅ debounced server `?search=` fetch; options from server not fixed-20; **stale-response race guarded**; empty-search clears (avoids re-introducing the bug); onSelect + homeAddress prefill + submission intact; helpers actually used |
| Scope / risk files | ✅ 3 files, +91/-9 net prod; none |
| **Revert-check** | ⚠️ **inconclusive for the .vue wiring** — a revert of only the component wiring (helper intact) wouldn't be caught. The committed test pins the extracted helper, not the reactive fetch. |

### Why flagged
Same policy as any fix whose committed test can't catch its own regression: it doesn't meet the auto-merge/deploy bar even though it's proven correct today. The un-testable part is the reactive wiring; by inspection it's correct and complete.

**Recommended action: merge it** — the fix is sound and the residual gap is frontend test infrastructure (a browser/component-test runner), not this PR. Tracked conceptually alongside the other harness gaps.

PR #72, opened 2026-07-10
Fix #62: reset shared e2e DB to seed snapshot before every test file

### LINKED ISSUE #62: E2E shared seeded DB is order-dependent — mutating tests can flake (harness robustness)

TusharW4ni:
## What was broken

The e2e harness (`tests/global-setup.ts`, `vitest.config.ts`) boots ONE Nuxt server against ONE seeded SQLite DB and runs `tests/e2e/*.test.ts` sequentially (`fileParallelism: false`). Mutating tests (create/complete/delete rides, clear phones, …) that don't fully clean up can corrupt seeded rows that later files assert against, making the suite **order-dependent** and prone to intermittent flakes — which would mislead the auto-merge gate.

## What changed (test harness only)

- `tests/utils/reset-db.ts` (new): `captureSeedSnapshot()` snapshots every user table's rows into `.data/test-seed-snapshot.json` (gitignored) once, right after the initial seed. `resetDb()` restores that snapshot in-process — a generic delete-all + bulk re-insert in a single transaction with `PRAGMA defer_foreign_keys = ON` over the same better-sqlite3 file the app uses — then clears cached login cookies (the wiped `session` rows).
- `tests/global-setup.ts`: calls `captureSeedSnapshot()` after seeding (before the app build).
- `tests/setup-reset-db.ts` (new) + `vitest.config.ts`: registered as a `setupFile`, so a `beforeAll(resetDb)` runs once per worker/file. Every test file now starts from an identical, freshly-seeded DB regardless of what earlier files did.
- `tests/utils/auth.ts`: adds `clearSessionCache()` (called by `resetDb`).

Chosen over re-running `tsx prisma/seed.ts` per file (~1.6s × 26 files ≈ +40s). The snapshot restore adds ~42ms/file (setup total ~1.2s across 28 files), so suite runtime stays ~28s.

## Verification

Regression tests (new): `tests/e2e/aaa-order-dependence-mutator.test.ts` deletes every seeded ride and does **not** clean up; `tests/e2e/zzz-order-dependence-victim.test.ts` (runs later, alphabetical) asserts the seed is still intact.

**Pre-fix** (confirmed by neutering `resetDb()` to a no-op to simulate `main`, then running the full suite): the mutator's deletions leaked forward and both the new victim file **and the pre-existing `smoke.test.ts`** failed with:

```
AssertionError: expected 0 to be greater than or equal to 9
 ❯ tests/e2e/zzz-order-dependence-victim.test.ts:30:26
     expect(rides.length).toBeGreaterThanOrEqual(9)
```

**Post-fix**: `pnpm test` → **Test Files 28 passed (28) / Tests 108 passed (108)**, ~28s.

Fixes #62

### DIFF
--- a/tests/e2e/aaa-order-dependence-mutator.test.ts
+++ b/tests/e2e/aaa-order-dependence-mutator.test.ts
@@ -1,0 +1,33 @@
+import { describe, expect, it } from 'vitest'
+import { $fetch } from '@nuxt/test-utils/e2e'
+import { bootShared } from '../utils/harness'
+import { loginAs } from '../utils/auth'
+
+// Order-independence regression, part 1 of 2 (issue #62).
+//
+// Vitest runs files alphabetically with fileParallelism:false, so this `aaa-`
+// file runs BEFORE `zzz-order-dependence-victim.test.ts`. It deliberately
+// destroys seeded rows (deletes every ride) and does NOT clean up after
+// itself — exactly the hostile, poorly-behaved mutating test the harness must
+// tolerate. Before the fix this corruption leaked into every later file; now
+// the per-file reset in tests/setup-reset-db.ts restores the seed snapshot at
+// the start of each file, so no file inherits these deletions.
+
+await bootShared()
+
+describe('order-dependence: hostile mutator', () => {
+  it('deletes every seeded ride and leaves the DB corrupted for the next file', async () => {
+    const cookie = await loginAs('reachtusharwani@gmail.com')
+
+    const before = await $fetch<{ id: string }[]>('/api/get/rides', { headers: { cookie } })
+    expect(before.length).toBeGreaterThanOrEqual(9)
+
+    for (const ride of before) {
+      await $fetch(`/api/delete/rides/${ride.id}`, { method: 'DELETE', headers: { cookie } })
+    }
+
+    const after = await $fetch<unknown[]>('/api/get/rides', { headers: { cookie } })
+    expect(after.length).toBe(0)
+    // Intentionally no cleanup: later files must not depend on our leftovers.
+  })
+})
--- a/tests/e2e/zzz-order-dependence-victim.test.ts
+++ b/tests/e2e/zzz-order-dependence-victim.test.ts
@@ -1,0 +1,36 @@
+import { describe, expect, it } from 'vitest'
+import { $fetch } from '@nuxt/test-utils/e2e'
+import { bootShared } from '../utils/harness'
+import { loginAs } from '../utils/auth'
+
+// Order-independence regression, part 2 of 2 (issue #62).
+//
+// This `zzz-` file runs AFTER `aaa-order-dependence-mutator.test.ts`, which
+// deletes every seeded ride and does not clean up. Its assertions depend on the
+// seeded rows still being present.
+//
+// PRE-FIX (no per-file reset): the mutator's deletions persist and this file
+// sees 0 rides — `toBeGreaterThanOrEqual(9)` fails, proving the suite is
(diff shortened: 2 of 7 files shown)


TusharW4ni:
## ✅ Automated verification — PASSED ALL AUTO-MERGE GATES (awaiting human merge)

Ran through the `/issue-loop` gate with an independent reviewer (fresh worktree, no author context) + a 3-angle `/review` pass. Result: **ready to merge**, but the merge was held for human approval (merging `main` deploys to prod, and this PR was authored+reviewed by automation only).

**Gate results:**
- **Suite (reviewer-run, 2×):** green, deterministic — 28 files / 108 tests, ~29s. Not trusting author's claim; re-run independently.
- **Revert-check:** PASS. With the fix reverted (harness files restored to `origin/main`, new files removed, regression tests kept), the victim fails `AssertionError: expected 0 to be greater than or equal to 9` at `zzz-order-dependence-victim.test.ts:26`; restoring the fix makes it pass. The regression genuinely pins the bug.
- **Scope:** 7 files, all under `tests/` + `vitest.config.ts`. Net non-test diff **+141 / -0**. No product code.
- **Risk files:** none (`vitest.config.ts` and `tests/utils/auth.ts` are not risk config; no `prisma/schema`, `.github/`, docker, deps, or server auth touched).
- **Correctness review:** reset mechanism is sound — generic table discovery covers all 10 user tables, `PRAGMA defer_foreign_keys=ON` handles the wipe/reload FK ordering, `clearSessionCache()` invalidates cookies whose `session` rows were reset, reset only runs in per-file `beforeAll` (can't mask mid-file failures).
- **A flagged "SQLITE_BUSY window" finding was refuted:** better-sqlite3 v12.6.2 defaults `busy_timeout=5000ms` (verified empirically), and the only fire-and-forget server path (`broadcastNotification`) is a read; the reset connection waits out any brief lock rather than throwing.

**Non-blocking follow-ups (optional, not gating):**
- Fragile implicit hook ordering: `reminder-claim-before-send.test.ts` inserts fixtures in its own `beforeAll`, relying on `setupFiles`' `resetDb` registering/running first. Holds today; worth a comment documenting the dependency.
- Minor cleanup: the unused `NUXT_TEST_DB_SNAPSHOT` env knob (YAGNI), passing the db path to `captureSeedSnapshot(path)` instead of the `process.env.DATABASE_URL` side-channel, and memoizing the snapshot read per worker.

_Automated `/issue-loop` verification. Merge + deploy-watch deferred to a human due to the production-deploy gate._

Issue #29, opened 2026-07-04
Production Database Dump with Real PII and Session Tokens in Public Repo


TusharW4ni:
### Location
`prod.sql` (repo root, tracked in git) — and `dev.db` in git history (added in `a6b8a88`, since deleted)

### Description
This repository is **public**, and `prod.sql` is a full dump of the production SQLite database. It currently contains:

- **41 real users** — names, email addresses, phone numbers
- **12 client profiles with 22 street addresses** — home addresses of the ride-program clients
- **20 rides** — who was driven where and when
- **10 auth sessions** — session tokens, IP addresses, and user agents

Anyone on the internet can read this. Even after the file is deleted, it remains fully visible in git history and in GitHub's commit cache. The session tokens in the dump have expired by now, but the PII does not expire.

### Remediation (in order)
1. **Remove the files from the working tree**: `git rm prod.sql` (and see the repo-hygiene issue for `deploy.tar.gz`). Add `*.sql` dumps and `*.db` to `.gitignore`.
2. **Purge history**: use `git filter-repo` (or BFG) to strip `prod.sql` and `dev.db` from all history, then force-push. All collaborators must re-clone.
3. **Purge GitHub's caches**: force-pushing does not remove cached commit views or forks. Open a GitHub Support ticket ("remove sensitive data") referencing the old commit SHAs, and check for forks.
4. **Rotate secrets**: rotate `BETTER_AUTH_SECRET` and truncate the `session` table in production so any dumped token lineage is dead.
5. **Report per org policy**: this is real client PII from a community-partner program — UTDallas EPICS and the partner org likely have notification expectations. Flag it to the team lead/mentor.
6. **Prevent recurrence**: if a prod snapshot is ever needed for debugging, use a scrubbed/anonymized copy, and share out-of-band — never through the repo.

### Why this exists
`prisma/migrations` is gitignored (see the migrations issue), so the dump was likely committed as an ad-hoc way to move schema+data around. Fixing the migrations workflow removes the temptation.

TusharW4ni:
🤖 **Skipped by `/issue-loop` — requires human action outside the repository.**

The in-repo step (`git rm prod.sql`, gitignore the pattern) is small, but the actual remediation cannot be automated safely:
- **git history rewrite** (`git filter-repo`/BFG) + force-push — destroys shared history, every collaborator must re-clone. Needs coordination.
- **GitHub Support ticket** to purge cached commit views / fork copies.
[...]

Issue #28, opened 2026-04-02
Audit Logging System


TusharW4ni:
### Description
For an app handling location data, scheduling, and client PII, an admin audit log is crucial for accountability and troubleshooting. We need a system that tracks exactly who did what (e.g., "Admin Tushar upgraded Alice to Volunteer", "Volunteer Bob cancelled their assignment for Ride #123").

### Implementation Details
- Create a new Prisma model: `AuditLog { id, userId, action, targetId, targetType, details (JSON), createdAt }`.
- Inject a utility function into critical API routes (`POST`, `PUT`, `DELETE`) to write a record to this table after a successful action.
- Build a simple Admin-only UI view (`/admin/audit`) to display and filter these logs.

Issue #14, opened 2026-04-02
[HIGH] Uncached Google Maps API Calls


TusharW4ni:
### Location
`server/api/get/rides/estimate/[id].ts`

### Description
Every time a user opens a ride detail page (`/rides/[id]`), the server makes a live HTTP request to the Google Maps Directions API. There is zero caching. This will rapidly drain API quotas and rack up large, unexpected billing costs, especially since the route for completed or scheduled rides rarely changes.

### Recommended Fix
Cache the distance/duration estimate locally on the `Ride` record in the database upon creation/update, falling back to the API only if the cached value is missing.

TusharW4ni:
🤖 **Skipped by `/issue-loop` — blocked on #33 (schema + migration).**

The fix caches the Maps distance/duration **on the Ride record**, which requires new columns (e.g. `cachedDistanceValue`, `cachedDurationValue`, `estimatedAt`) — a `prisma/schema/` change and therefore a production migration, which is blocked on #33 (migrations are gitignored; no migrate-deploy path). The gate always routes schema changes to human review, and #33 must be resolved first for a safe prod rollout.

Also note the benefit path (the Google Directions call) isn't exercisable in the e2e harness (no API key/network), so verification would rely on caching-layer logic tests plus a manual check. Best done once #33 lands. Happy to prep the caching logic + schema once migrations are unblocked.

PR #76, opened 2026-07-10
Fix #33: adopt a Prisma migrations workflow

### LINKED ISSUE #33: Prisma Migrations Are Gitignored — No Reproducible Schema History

TusharW4ni:
Fixes #33.

Adopts a real Prisma migrations workflow so the schema has reproducible history in version control and prod can be advanced with `prisma migrate deploy` — unblocking #14, #27, #28, #5.

## ⚠️ Human review + one manual prod step required (do NOT auto-merge)
This PR touches **risk files** (`prisma/migrations`, `dockerfile`, `package.json`) and changes the deploy process. Before/right-after merging you must run **once** against the **live prod DB** (which already has this schema from the old `db push` process):

```
DATABASE_URL="<prod db url>" npx prisma migrate resolve --applied 20260710023845_init
```

This marks the baseline as already-applied **without re-running it**, so the entrypoint's `prisma migrate deploy` sees the DB as up-to-date and is a no-op. Skipping this step would make `migrate deploy` try to CREATE tables that already exist and fail. After that one-time resolve, all future schema changes ship as new migrations applied automatically at container start.

## What changed
- **`.gitignore`** — removed `/prisma/migrations`; added `!prisma/migrations/**/*.sql` to carve migration SQL out of the `*.sql` dump-block from #29 (so migrations are tracked, but `prod.sql`-style dumps stay ignored — verified both).
- **`prisma/migrations/20260710023845_init/`** — squashed baseline migration (all 10 tables + indexes) generated from the current `prisma/schema/`.
- **`dockerfile`** — final stage now also copies `prisma.config.ts`, installs the pinned `prisma` CLI (+`dotenv`, which the config loads) into the app root, and copies the entrypoint. `CMD` is now the entrypoint.
- **`docker-entrypoint.sh`** (new) — runs `prisma migrate deploy` (via `npx --no-install`, so no runtime network fetch), then `exec node .output/server/index.mjs`. `set -e` fails fast if a migration errors.
- **`package.json`** — replaced the stale `prisma:reset` (which regenerated an `init` migration) with `prisma:migrate` (`migrate dev`), `prisma:deploy` (`migrate deploy`), and a migration-aware `prisma:reset` (`migrate reset --force`). **No dependency changes.**
- **`CLAUDE.md`** — documented the new workflow and the `.gitignore` carve-out.

## Verification done
- ✅ **Baseline schema is byte-identical to `db push`**: created one DB via `prisma db push` and one via `prisma migrate deploy`, diffed `.schema` — the 105 application-schema lines match exactly; the only delta is Prisma's own `_prisma_migrations` bookkeeping table (expected). This is what makes the `migrate resolve --applied` step safe on prod.
- ✅ **Seed works** against a `migrate deploy` DB.
- ✅ **`migrate deploy` is idempotent** — a second run reports "No pending migrations to apply", so container restarts are safe.
- ✅ **Runtime path validated in isolation** — a clean dir with only `prisma/` + `prisma.config.ts` + a freshly-installed `prisma` CLI runs `migrate deploy` successfully (mirrors what the final image does).
- ✅ **e2e suite green** — 29 files / 111 tests; the harness still uses `db push` and is unaffected by the committed migrations.

## Not validated here (please confirm in CI)
- 🔶 **The Docker image build was not run locally** (the dev sandbox has no container-registry network, so `node:22-slim` couldn't be pulled). The Dockerfile changes are standard COPY/RUN/CMD steps and the migrate-deploy logic they invoke is validated above, but the actual `linux/arm64` image build + a container-start smoke test should be confirmed by the CI build (and ideally a staging run) before this reaches prod. Because merging `main` builds+pushes the prod image, consider building the image from this branch first if you want belt-and-suspenders.

## Notes
- `migrate deploy` at **container start** (not CI) is deliberate: the prod DB is a runtime SQLite file (persistent volume), which CI's build environment can't reach.
- Image size grows modestly (prisma CLI + engine in the final stage) — the cost of runtime migrations.


TusharW4ni:
## 🔀 Merge conflict resolved

Merged `origin/main` (which now includes #75 repo-hygiene) into this branch. Two conflicts, both resolved cleanly and validated:

- **`.gitignore`** — kept BOTH sides: #75's `.gemini/` ignore rule AND this branch's `!prisma/migrations/**/*.sql` carve-out (so migration SQL stays tracked while dumps + `.gemini/` stay ignored). `/prisma/migrations` remains un-ignored.
- **`server/utils/auth.ts`** — auto-merged correctly: #75's removal of the dead `nodemailer` import + `transporter` const is kept, AND the soft-delete login-block (`findFirst({ email, deletedAt: null })`) is preserved. No overlap.
- #75's file deletions (`fix-rides.ts`, `PLAN.md`, `.gemini/settings.json`) carried through.

Validated the merged result: `pnpm test` → **30 files / 116 tests passing**. Branch is now mergeable.

_Note: this branch now carries #33 (migrations) **and** #27 (soft deletes), since #79 was merged into this base branch. PRs #77 (#5) and #78 (#14) are still based on the earlier tip and will retarget to `main` once this merges — they may need a trivial rebase._

PR #77, opened 2026-07-10
Fix #5: build real ride cancellation (CANCELLED status)

### LINKED ISSUE #5: [HIGH] Impossible State / Schema Violation on Ride Cancellation
### Location
`server/api/put/rides/[id].ts` (Lines 85-88)

### Description
The code contains logic to handle `body.status === 'CANCELLED'`. However, in `prisma/schema.prisma`, `RideStatus` is strictly enforced via an enum containing only `CREATED`, `ASSIGNED`, and `COMPLETED`. Attempting to pass `CANCELLED` here will result in Prisma throwing a hard validation exception.

### Recommended Fix
Either add `CANCELLED` to the Prisma `RideStatus` enum or handle cancellations structurally via a separate boolean flag (e.g., `isCancelled`).

### LINKED ISSUE #22: [UI/Logic] Un-toggleable 'Cancelled' Status Filter in Rides Dashboard

TusharW4ni:
## What was broken
`CANCELLED` was referenced throughout the code (notification type, seed template, settings toggle, a `RIDE_CANCELLED` branch in the ride PUT handler) but was **not** a member of the `RideStatus` enum, and after #31 the PUT validation only allowed `CREATED|ASSIGNED|COMPLETED`. So `PUT /api/put/rides/:id {status:'CANCELLED'}` returned **400** and the `RIDE_CANCELLED` notification branch was dead/unreachable. Per the owner's decision on #5, this builds a **real cancellation feature** (Option B), not dead-code removal.

## What changed
- **`prisma/schema/enum.prisma`** — add `CANCELLED` to `RideStatus`. *(risk file — schema; see migration note below.)*
- **`server/api/put/rides/[id].ts`** — allow `CANCELLED` in the #31 zod whitelist and the `updateData` type, so cancellation returns 200. This makes the existing `RIDE_CANCELLED` branch reachable; it fires `sendNotification('RIDE_CANCELLED', ...)` for the volunteer who was assigned before cancellation. Removed the obsolete `as string` casts and the stale "no CANCELLED status" comment.
- **`app/pages/rides/[id].vue`** — admin-gated **Cancel Ride** button + confirmation modal (mirrors the Delete flow), hidden once a ride is already `CANCELLED`/`COMPLETED`; `CANCELLED` gets an `error`-colored status badge.
- **`app/pages/rides/index.vue` + `app/utils/rideFilters.ts`** — `CANCELLED` is now a real, toggleable status filter and a default exclusion (finished work), with an `error` badge in the table.

## Migration note
SQLite stores Prisma enums as `TEXT` (the #33 init baseline already defines `status TEXT NOT NULL DEFAULT 'CREATED'` with no CHECK constraint), so adding an enum value generates **no SQL migration**. `prisma migrate dev --name add_cancelled_ride_status` reported the schema already in sync and created **no new migration file** — expected. No production migration is required for this change.

## Stacked on #76 (issue #33)
This PR is **stacked on `fix/33-adopt-migrations` (#76)** and targets it as the base branch so the diff shows only the cancellation delta. **It must merge after #76.**

## Verification
- **Regression test** `tests/e2e/ride-cancel.test.ts`:
  - As ADMIN, `PUT /api/put/rides/:id {status:'CANCELLED'}` → asserts **200** and GET-back persists `status === 'CANCELLED'`.
  - Cancels an `ASSIGNED` ride and asserts the volunteer is retained (exercising the `RIDE_CANCELLED` notification path — the email send is swallowed in test env, and the server log shows the attempted send).
  - **Pre-fix FAIL** (observed): `AssertionError: expected 400 to be 200` and `FetchError: [PUT] ... 400 Invalid request body`.
- Updated `tests/e2e/ride-filters.test.ts` (#22): the case that previously asserted `status:CANCELLED` was stripped as *stale* now asserts it is a valid, retained filter (CANCELLED is a real status as of this PR); added a separate case using a genuinely unknown value to keep sanitize coverage. The #22 invariant test (every default exclusion is toggleable) still passes.
- **Full suite:** `pnpm test` → **30 files / 114 tests passed**. `pnpm build` succeeds.

## Scope / risk
Touched files: `prisma/schema/enum.prisma` (schema — **flagged for human review**), the PUT rides route, the ride details + index pages, the shared ride-filters helper, and tests. No changes to auth, CI, dockerfile, or deps. Net non-test change: +68/-11 lines.

Fixes #5

<sub>Not closing #22: the stuck-filter bug it describes was already fixed on the base branch (`rideFilters.ts` sanitize). This PR keeps that fix coherent now that CANCELLED is real, but does not itself resolve #22.</sub>

### DIFF
--- a/app/pages/rides/[id].vue
+++ b/app/pages/rides/[id].vue
@@ -14,6 +14,7 @@
   const isEditModalOpen = ref(false)
   const isCompleteModalOpen = ref(false)
   const isDeleteModalOpen = ref(false)
+  const isCancelModalOpen = ref(false)
 
   const isAdmin = computed(() => session.value?.user?.role === 'ADMIN')
   const isVolunteer = computed(() => session.value?.user?.role === 'VOLUNTEER')
@@ -138,6 +139,20 @@
     }
   }
 
+  async function handleCancel() {
+    try {
+      await $fetch(`/api/put/rides/${id}`, {
+        method: 'PUT',
+        body: { status: 'CANCELLED' },
+      })
+      toast.add({ title: 'Success', description: 'Ride cancelled', color: 'success' })
+      isCancelModalOpen.value = false
+      await refreshRide()
+    } catch (err) {
+      toast.add({ title: 'Error', description: 'Failed to cancel ride', color: 'error' })
+    }
+  }
+
   async function handleVolunteerAction(action: 'signup' | 'unsignup' | 'complete') {
     if (action === 'signup') {
       try {
@@ -222,7 +237,16 @@
           <template #header>
             <div class="flex items-center justify-between">
               <h2 class="text-xl font-bold">Ride Information</h2>
-              <UBadge :color="ride.status === 'COMPLETED' ? 'success' : 'info'" variant="subtle">
+              <UBadge
+                :color="
+                  ride.status === 'COMPLETED'
+                    ? 'success'
+                    : ride.status === 'CANCELLED'
+                      ? 'error'
+                      : 'info'
+                "
+                variant="subtle"
+              >
                 {{ ride.status }}
               </UBadge>
             </div>
@@ -285,6 +309,14 @@
                 variant="subtle"
                 class="flex-1 justify-center"
                 @click="isEditModalOpen = true"
+              />
+              <UButton
+                v-if="isAdmin && ride.status !== 'CANCELLED' && ride.status !== 'COMPLETED'"
+                label="Cancel Ride"
+                icon="i-lucide-ban"
+                color="warning"
+                variant="subtle"
+                @click="isCancelModalOpen = true"
               />
               <UButton
                 v-if="isAdmin"
@@ -524,6 +556,27 @@
               @click="isDeleteModalOpen = false"
             />
             <UButton label="Delete" color="error" @click="handleDelete" />
+          </div>
+        </div>
+      </template>
+    </UModal>
+
+    <!-- Cancel Ride Confirmation Modal -->
+    <UModal v-model:open="isCancelModalOpen" title="Cancel Ride">
+      <template #content>
+        <div class="space-y-4 p-4">
+          <p>
+            Are you sure you want to cancel this ride? The assigned volunteer (if any) will be
+            notified.
+          </p>
+          <div class="flex justify-end gap-2">
+            <UButton
+              label="Keep Ride"
+              color="neutral"
+              variant="ghost"
+              @click="isCancelModalOpen = false"
+            />
+            <UButton label="Cancel Ride" color="warning" @click="handleCancel" />
           </div>
         </div>
       </template>
--- a/app/pages/rides/index.vue
+++ b/app/pages/rides/index.vue
@@ -297,6 +297,7 @@
             CREATED: 'info' as const,
             ASSIGNED: 'warning' as const,
             COMPLETED: 'success' as const,
+            CANCELLED: 'error' as const,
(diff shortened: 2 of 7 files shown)


TusharW4ni:
## 🚩 FLAGGED for human review — schema risk file (feature is correct & verified)

Built per the owner's Option-B decision on #5. Ran the full `/issue-loop` verification (independent reviewer in a fresh worktree + an enum-exhaustiveness `/review` pass). **Every gate passes except the mandatory rule that a `prisma/schema/*` change needs human review.** No blocker-severity findings.

**Stacked on #76** (base = `fix/33-adopt-migrations`) — merge after #76.

**Gate results:**
- **Suite (reviewer-run):** green — 30 files / 114 tests, ~29s.
- **Revert-check:** PASS. Reverting the 5 fix files to base makes `ride-cancel.test.ts` fail (`CANCELLED` rejected → 400 instead of 200); restoring makes it pass. Genuine regression test.
- **No migration file** (correct): SQLite stores Prisma enums as TEXT, so adding `CANCELLED` is not a DDL change — `prisma/migrations/` is unchanged vs base.
- **Scope:** 7 files, net non-test **+57**. **Only risk file: `prisma/schema/enum.prisma`.** No CI/docker/deps/auth.
- **Enum-exhaustiveness sweep (the main risk of a 4th status):** metrics `topRiders`/`hours` filter `COMPLETED` (unaffected); reminders cron selects `ASSIGNED` only (cancelled rides get no reminders); signup/unsignup guard `CREATED`/`ASSIGNED` (can't sign up for a cancelled ride); the `RIDE_CANCELLED` notification fires only on a real transition and no-ops safely with no volunteer; filters are coherent and toggleable (not a #22 regression). Badge colors updated in both list + detail views.

**Two decisions for you (both non-blocking):**
1. **`completionRate` denominator** (`server/api/get/metrics/completionRate/index.ts`) — `totalRides` counts *all* rides while `completedRides` counts `COMPLETED`, so cancelled rides now lower the completion %. Reviewers split on whether this is desired (a cancelled ride genuinely wasn't completed) or should be excluded. **Recommendation: exclude CANCELLED** (`status: { not: 'CANCELLED' }`) so the metric means "of rides that could complete." One-line change — tell me and I'll add it, or adjust on review.
2. **No server-side status-transition guard** — the API accepts any transition (e.g. `COMPLETED→CANCELLED`, `CANCELLED→CREATED`); the UI hides the button but the server doesn't enforce. This is **pre-existing** (the route never validated transitions) and out of #5's scope — worth a separate data-integrity issue if you want a state machine.

_Automated `/issue-loop` verification. Flagged per the risk-file rule; feature verified correct and safe._

PR #82, opened 2026-07-13
Fix #57: dedup addresses on a matchKey, preserve original display casing

### LINKED ISSUE #57: Address Title-Case normalization mangles acronyms/directionals in stored display (#16 polish)
### Follow-up from #16 (PR #56)

#16 normalizes addresses (trim + collapse whitespace + Title-Case) before upsert to dedup case/whitespace variants. Correct for dedup, but Title-Case degrades the canonical `Address` row that backs `Client.homeAddress` (shown on the people page) for a minority of inputs:
- `"123 NW 5th Ave"` → `"123 Nw 5th Ave"`
- `"PO Box 12"` → `"Po Box 12"`
- `"MLK Jr Blvd"` → `"Mlk Jr Blvd"`
- `"McDonald Ave"` → `"Mcdonald Ave"`

Dedup is unaffected; this is display-only cosmetics.

### Proper fix (deferred by design constraint)
The clean solution — dedup on a normalized key while preserving the user's original casing for display — needs a separate normalized column on `Address` (e.g. `matchKey`), which is a schema change + migration, currently blocked on #33 (migrations gitignored). Once #33 is resolved: add a normalized/uniqued `matchKey` derived from the fields, drop the unique on the raw fields, and store the original-cased street/city for display.

Low priority; the current Title-Case behavior is an acceptable interim.

TusharW4ni:
## What was broken

#16 (PR #56) normalized addresses by **Title-Casing** every field before upserting on the raw `@@unique([street, city, state, zip])` constraint. That deduped case/whitespace variants correctly, but Title-Case **mangled the stored display value** for a minority of inputs — acronyms, directionals, and mixed-case — and that canonical `Address` row backs `Client.homeAddress` shown on the people page:

- `"123 NW 5th Ave"` → `"123 Nw 5th Ave"`
- `"PO Box 12"` → `"Po Box 12"`
- `"MLK Jr Blvd"` → `"Mlk Jr Blvd"`
- `"McDonald Ave"` → `"Mcdonald Ave"`

## What changed

Split dedup from display (the issue's "proper fix", unblocked now that #33 landed the migrations workflow):

- **Schema** (`prisma/schema/client.prisma`): added `matchKey String @unique`, **dropped** `@@unique([street, city, state, zip])`. `street/city/state/zip` now store the user's **original casing**.
- **`server/utils/address.ts`**: `normalizeAddress` now returns display fields (trim + internal-whitespace-collapse only — **no Title-Case, no lowercase**) plus a `matchKey` (`addressMatchKey`: lowercased + collapsed fields joined on `|`). `matchKey` is exported so the migration backfill can reproduce it.
- **Upsert sites** (`post/clients`, `put/clients/[id]`, `post/rides`): dedup on `where: { matchKey }`, write the original-cased display fields.
- **`prisma/seed.ts`**: builds addresses (incl. `matchKey`) through the same util.

## ⚠️ Migration — needs human review (schema + constraint change + backfill)

`prisma/migrations/20260713230446_address_matchkey/migration.sql` is **NOT purely additive**. It adds a `NOT NULL matchKey` and swaps the unique constraint, so on a **populated prod DB** it:

1. **Backfills `matchKey`** for every existing row from the stored fields: `lower(trim(street))||'|'||lower(trim(city))||'|'||lower(trim(state))||'|'||lower(trim(zip))`. Rows written under #16 are already whitespace-collapsed and Title-Cased, and `lower()` erases the Title-Case, so the backfilled key equals what a fresh upsert of the same address now computes.
2. **De-dupes before adding `UNIQUE(matchKey)`**: any pre-existing rows that collapse to the same key (legacy duplicates predating #16) would violate the new index, so the migration repoints every FK (`client.homeAddressId`, `ride.pickupAddressId`, `ride.dropoffAddressId`) onto the surviving row (`MIN(id)` per key) and keeps only that row. On a post-#16 DB the rows are already case-folded-unique, so this step is a no-op. **Collision risk on real data is low** and handled defensively.

**Known limitation (documented in the migration):** a legacy row with *un-collapsed internal whitespace* (pre-#16) backfills a key that a future JS upsert of the same address (which collapses internal whitespace) won't match — worst case a benign re-duplicate row, never a failure.

Verified: `prisma migrate deploy` applies cleanly on a fresh DB; the migration is consistent with the schema (`prisma migrate status` → up to date); and a scratch-DB test with an acronym row + a legacy duplicate pair + FKs confirmed the acronym display is preserved verbatim, the dupes collapse to one row, and both client and ride FKs repoint to the survivor.

## Verification

- **`tests/e2e/address-matchkey-display.test.ts`** (new):
  - *"stores the ORIGINAL casing of an acronym/directional street for display"* — creates a client with `"123 NW …th Ave"`, reads it back via `/api/get/clients`, asserts the stored street equals the input.
  - *"still dedups case/whitespace variants onto a single Address row"* — creates a ride pickup with the same address in different casing + messy whitespace and asserts it resolves to the client's existing `homeAddressId`.
- **Pre-fix failing assertion** (captured by stashing the source fix and running the test against the old Title-Case code):
  > `AssertionError: expected '123 Nw 242499549th Ave' to be '123 NW 242499549th Ave'`
  The dedup test **passed** both pre- and post-fix (dedup was never broken).
- **Full suite:** `pnpm test` → **33 files, 123 tests passed** (the SMTP errors in output are the pre-existing swallowed send failures, #25).

## Risk files
- `prisma/schema/client.prisma` + a new migration with a **backfill + unique-constraint swap + FK repoint** → **human review mandatory**.
- No CI/docker/deps/auth changes.

Fixes #57

### DIFF
--- a/prisma/migrations/20260713230446_address_matchkey/migration.sql
+++ b/prisma/migrations/20260713230446_address_matchkey/migration.sql
@@ -1,0 +1,78 @@
+-- Address dedup key (issue #57).
+--
+-- Splits dedup from display: `matchKey` is a lossy, case/whitespace-insensitive
+-- key (uniqued) while street/city/state/zip now preserve the user's ORIGINAL
+-- casing. Replaces the old UNIQUE(street, city, state, zip), whose enforcement
+-- relied on Title-Casing the display fields and mangled acronyms/directionals
+-- ("123 NW 5th Ave" -> "123 Nw 5th Ave", "PO Box 12" -> "Po Box 12").
+--
+-- PRODUCTION DATA HANDLING -- this is NOT a purely-additive migration:
+--   1. `matchKey` is NOT NULL with no default, so it is BACKFILLED for every
+--      existing row from the stored fields:
+--        lower(trim(street))||'|'||lower(trim(city))||'|'||lower(trim(state))||'|'||lower(trim(zip))
+--      This reproduces the key the app now computes (server/utils/address.ts:
+--      addressMatchKey). Rows written under #16 are already whitespace-collapsed
+--      and Title-Cased, and lower() erases the Title-Case, so the backfilled key
+--      matches what a fresh upsert of the same address would produce.
+--   2. The new UNIQUE(matchKey) replaces UNIQUE(street, city, state, zip). Any
+--      pre-existing rows that collapse to the SAME matchKey (legacy duplicates
+--      predating #16) would violate it, so we DE-DUPE first: repoint every FK
+--      (client.homeAddressId, ride.pickupAddressId, ride.dropoffAddressId) onto
+--      the surviving row -- the lowest id per matchKey -- and keep only that row.
+--      On a post-#16 DB the rows are already case-folded-unique, so the de-dupe
+--      and FK repoint are no-ops. Collision risk on real data is therefore low;
+--      it is handled here purely to keep the migration safe on any older dump.
+PRAGMA defer_foreign_keys=ON;
+PRAGMA foreign_keys=OFF;
+
+-- 1. Repoint FKs from duplicate addresses onto the survivor (MIN id per key),
+--    so dropping the duplicate rows below cannot orphan a client or ride.
+UPDATE "client"
+SET "homeAddressId" = (
+    SELECT MIN(a2."id") FROM "address" a2
+    WHERE lower(trim(a2."street"))||'|'||lower(trim(a2."city"))||'|'||lower(trim(a2."state"))||'|'||lower(trim(a2."zip"))
+        = (SELECT lower(trim(a1."street"))||'|'||lower(trim(a1."city"))||'|'||lower(trim(a1."state"))||'|'||lower(trim(a1."zip"))
+           FROM "address" a1 WHERE a1."id" = "client"."homeAddressId")
+);
+
+UPDATE "ride"
+SET "pickupAddressId" = (
+    SELECT MIN(a2."id") FROM "address" a2
+    WHERE lower(trim(a2."street"))||'|'||lower(trim(a2."city"))||'|'||lower(trim(a2."state"))||'|'||lower(trim(a2."zip"))
+        = (SELECT lower(trim(a1."street"))||'|'||lower(trim(a1."city"))||'|'||lower(trim(a1."state"))||'|'||lower(trim(a1."zip"))
+           FROM "address" a1 WHERE a1."id" = "ride"."pickupAddressId")
+)
+WHERE "pickupAddressId" IS NOT NULL;
+
+UPDATE "ride"
+SET "dropoffAddressId" = (
+    SELECT MIN(a2."id") FROM "address" a2
+    WHERE lower(trim(a2."street"))||'|'||lower(trim(a2."city"))||'|'||lower(trim(a2."state"))||'|'||lower(trim(a2."zip"))
+        = (SELECT lower(trim(a1."street"))||'|'||lower(trim(a1."city"))||'|'||lower(trim(a1."state"))||'|'||lower(trim(a1."zip"))
+           FROM "address" a1 WHERE a1."id" = "ride"."dropoffAddressId")
+)
+WHERE "dropoffAddressId" IS NOT NULL;
+
+-- 2. Rebuild the table with matchKey; keep one backfilled row per matchKey.
+CREATE TABLE "new_address" (
+    "id" TEXT NOT NULL PRIMARY KEY,
+    "street" TEXT NOT NULL,
+    "city" TEXT NOT NULL,
+    "state" TEXT NOT NULL,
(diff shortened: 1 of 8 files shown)


TusharW4ni:
## 🚩 FLAGGED for human review — schema + hand-written migration (fix verified correct & safe)

Full `/issue-loop` verification: independent reviewer (fresh worktree, incl. a **live migration test on a scratch DB with legacy duplicates + FKs**) + a migration/dedup `/review` pass. **Flag is mandatory** (`prisma/schema/*` change + a data-touching migration). No data-loss / orphaning / dedup-regression defect found.

**Gate results:**
- **Suite (reviewer-run):** green — 33 files / 123 tests.
- **Revert-check:** PASS. Reintroducing Title-Case in the display path makes `address-matchkey-display.test.ts` fail (`expected '123 Nw …' to be '123 NW …'`); restoring passes. Dedup test passes both ways (dedup never broken).
- **Migration (live-tested on a populated scratch DB):** succeeds; legacy duplicates collapse to one row per matchKey; **FKs repointed to the survivor with `foreign_key_check` empty (no orphans)**; acronym display preserved verbatim; `UNIQUE(matchKey)` present, old composite unique gone; full 4-migration chain deploys cleanly and matches the schema. FK-repoint runs before the row drop; `defer_foreign_keys=ON` handles integrity.
- **matchKey consistency:** identical for all app-written rows (every stored field is already whitespace-collapsed by `normalizeAddress`, so SQL `lower(trim())` == util `collapse().toLowerCase()`).
- **Upsert sites / seed:** all three upserts now key on `{ matchKey }` and store collapse-only original-case display; no caller uses the dropped composite key; seed builds via the same util.
- **Scope:** 8 files, net non-test **+112**. Risk files: `prisma/schema/client.prisma` + migration. No CI/docker/deps/auth.

**Findings for your decision (none are data-loss; all documented):**
1. **Backfill vs app-util divergence on LEGACY data (recommend deciding before prod deploy):** the SQL backfill uses `lower(trim(x))`, but the app computes `collapse(x).toLowerCase()` (collapse = trim + collapse *internal* whitespace; JS lower is full-Unicode vs SQLite `lower()` ASCII-only). For app-written rows these coincide (fields are pre-collapsed), but a **pre-#16 legacy row** with internal double-spaces/tabs, or any uppercase non-ASCII address, could backfill a key a later app upsert won't match → a **benign duplicate Address row** (never a constraint failure/orphan). Since the historical prod dump predates #16, if prod has such rows, prefer backfilling matchKey via the app util (a one-off script using `addressMatchKey`) or collapsing internal whitespace in the SQL. Documented in the migration header.
2. **Migration has no automated coverage:** the e2e harness provisions via `db push`, not `migrate deploy`, so the backfill/dedupe/FK-repoint SQL runs only in prod (I validated it manually on a scratch DB). Inherent to the harness.
3. **`state` no longer upper-cased (decision):** #16 upper-cased state (`tx`→`TX`); this fix collapse-only, so `state` now displays as typed. Either is defensible (postal-convention uppercase vs preserve-as-typed) — uppercasing `state` for display is safe (no acronym-mangling risk) if you prefer convention.
4. **Existing #16 rows keep their Title-Cased display** until an admin re-saves them — the fix is forward-looking (per the issue), not a backfill of old display values.

_Automated `/issue-loop` verification. Flagged per the risk-file policy; fix + migration verified correct and safe. Tell me and I'll apply the finding-#1 backfill hardening and/or the state-uppercase tweak._

TusharW4ni:
## Review findings resolved (sensible decisions applied)

- **Applied — `state` now upper-cased for display.** #16 upper-cased state; the initial fix dropped it, so `tx` displayed as `tx`. State is a 2-letter postal code (no acronym-mangling risk), so it's now upper-cased for display while street/city keep their original casing. `matchKey` lowercases regardless, so dedup is unaffected. Added a regression assertion. Suite green **33 files / 124 tests**.
- **Left as-is (deliberate) — backfill vs app-util divergence (finding #1).** The app computes `matchKey` from the already-collapsed display fields, so `lower(trim())` matches the app key for **every app-written row** — the divergence only affects pre-#16 legacy rows with exotic internal whitespace or uppercase non-ASCII, where the worst case is a benign duplicate row (no data loss/orphaning), documented in the migration header. Rewriting the collapse across the migration's 5 key-computing subqueries would add more risk to an already-verified-correct migration than it removes. If the prod dump contains such rows, a one-off app-side `addressMatchKey` backfill is the clean remedy — I can prep that separately.
- **Left as-is — no automated migration coverage (finding #2):** inherent to the harness (`db push`, not `migrate deploy`); validated manually on a populated scratch DB.
- **Left as-is — existing #16 rows keep their mangled display until re-saved (finding #4):** forward-looking by design, per the issue.

Still flagged `needs-human-review` for the schema + migration risk files. No migration SQL was touched by this update.

PR #84, opened 2026-07-14
Fix #30: split Google Maps key into server-only + public embed keys

### LINKED ISSUE #30: Google Maps API Key Exposed in Client Bundle via Public Runtime Config
### Location
`nuxt.config.ts` (`runtimeConfig.public.googleMapsApiKey`), used in `server/api/get/rides/estimate/[id].ts` and `app/pages/rides/[id].vue`

### Description
The Google Maps API key lives in **public** runtime config, so it ships in the client JavaScript bundle where anyone can extract it. The same key is used two different ways:

1. **Client-side** — the Maps Embed iframe in `app/pages/rides/[id].vue`. Public exposure is expected here, but only safe if the key is referrer-restricted.
2. **Server-side** — the Directions API call in `server/api/get/rides/estimate/[id].ts`. Server calls send no referrer, so this only works if the key is **unrestricted** — which means the key currently in the browser bundle is an unrestricted key. Anyone who extracts it can bill Directions/Places/etc. usage to the project's Google Cloud account.

### Recommended Fix
Split into two keys:
- `runtimeConfig.googleMapsServerKey` (private, no `public.`) for the Directions call — restrict it by IP (or at minimum by API) in Google Cloud Console.
- `runtimeConfig.public.googleMapsEmbedKey` for the iframe — restrict it to the production domain referrer and to the Maps Embed API only.

Then **rotate the current key**, since the unrestricted one must be assumed harvested (public repo + public bundle). Set daily quota caps on both keys as a backstop. Related: #14 (caching estimates) would also cut how often the server key is used.

TusharW4ni:
## What was broken

The Google Maps API key was exposed in the client bundle via **public** runtime config (`runtimeConfig.public.googleMapsApiKey`). The server-side estimate endpoint (`server/api/get/rides/estimate/[id].ts`) read that *public* key to call the Directions API even though it runs server-side, which meant the powerful Directions-capable key was shipped to the browser where anyone could scrape it and burn quota/billing.

## What changed (app-side split only — no key rotation)

- **`nuxt.config.ts`** (⚠️ **RISK FILE — human review mandatory**): added a private, server-only runtime config key alongside the existing public one:
  ```ts
  runtimeConfig: {
    googleMapsApiKey: '',            // server-only, NUXT_GOOGLE_MAPS_API_KEY
    public: {
      googleMapsApiKey: '',          // client embed, NUXT_PUBLIC_GOOGLE_MAPS_API_KEY
    },
  }
  ```
  Nuxt maps `NUXT_GOOGLE_MAPS_API_KEY` → `runtimeConfig.googleMapsApiKey` and `NUXT_PUBLIC_GOOGLE_MAPS_API_KEY` → `runtimeConfig.public.googleMapsApiKey`. Only `public.*` is serialized into the client bundle, so the server-only key never reaches the browser.
- **`server/api/get/rides/estimate/[id].ts`**: reads `useRuntimeConfig(event).googleMapsApiKey` (server-only) for the Directions call instead of `config.public.googleMapsApiKey`. The graceful `error: 'Maps API Key not configured'` fallback is preserved when the key is empty.
- **`app/pages/rides/[id].vue`**: unchanged behavior — the map-embed iframe keeps reading the **public** key (a client-visible embed key is correct by design). Added a code comment.
- **`.env.example` / `README.md`**: document both `NUXT_GOOGLE_MAPS_API_KEY` (server, Directions) and `NUXT_PUBLIC_GOOGLE_MAPS_API_KEY` (client, embed — must be referrer-restricted).

## Verification

**Regression test:** `tests/e2e/maps-key-split.test.ts` — *"estimate endpoint reads the empty SERVER key, not the set public key."* The harness (`tests/global-setup.ts`) sets **only** the bogus `NUXT_PUBLIC_GOOGLE_MAPS_API_KEY` and leaves the server key **unset**, then the test asserts a cache-miss estimate returns exactly `'Maps API Key not configured'`.

- **Post-fix:** the endpoint reads the *empty server* key → short-circuits at the "not configured" gate → returns the expected string and makes **no outbound call** (the green suite stays fully hermetic / zero egress).
- **Pre-fix failing assertion** (observed by temporarily reverting the endpoint to `config.public.googleMapsApiKey`): the endpoint reads the *set public* key, gets past the gate, and attempts a live Directions call:
  ```
  AssertionError: expected 'REQUEST_DENIED' to be 'Maps API Key not configured'
   ❯ tests/e2e/maps-key-split.test.ts:80:23
  ```

**Why the harness was inverted (review follow-up):** an earlier revision set the bogus key on the *server* var, which made cache-miss estimates fire a real outbound HTTPS call. On a blackhole-firewall CI (silent SYN drop) ofetch has no timeout, so the connect would hang until the 30s `testTimeout` aborted → suite RED purely from network posture. Setting the bogus value on the *public* var instead keeps the previously-hermetic suite hermetic and lets the exact assertion be used.

`tests/e2e/estimate-cache.test.ts` is unchanged in behavior — its `expect(miss.error).toBe('Maps API Key not configured')` assertion still holds (server key empty → deterministic miss, no network).

Also confirmed by grep: no `config.public.googleMapsApiKey` read remains anywhere in `server/`.

**Suite:** `pnpm test` → 35 files, 130 tests passed, zero outbound Google calls. **Build:** `pnpm build` → success.

## Deployment note (breaking config change for the operator)

Because the estimate endpoint now reads a **separate** server variable, a live deployment that only sets `NUXT_PUBLIC_GOOGLE_MAPS_API_KEY` will get `'Maps API Key not configured'` from the estimate feature until the operator **also** sets `NUXT_GOOGLE_MAPS_API_KEY`. This is intended, but the owner must set the new server var (alongside key rotation) or the distance/duration estimate silently stops working.

## Residual OWNER action (out of scope — Google Cloud Console)

This split only stops the powerful key from shipping to the client; it does **not** rotate or restrict anything in Google Cloud. The owner must:
1. Provision **two different** Google Cloud API keys.
2. Restrict the **server** Directions key by IP/API.
3. **HTTP-referrer-restrict** the **public** embed key and scope it to the Maps Embed API.
4. Assume the currently-exposed key is compromised (public repo + public bundle) and rotate it.

Fixes #30

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -6,4 +6,11 @@ BETTER_AUTH_URL=http://localhost:3000
 # TRUSTED_ORIGINS=https://rides.example.org
 EMAIL_USER=
 EMAIL_PASS=""
+# Google Maps keys (issue #30) — use TWO DIFFERENT Google Cloud keys:
+#  - NUXT_GOOGLE_MAPS_API_KEY: server-only, powers the Directions API distance
+#    estimate. Never ships to the browser. Restrict it by IP / API in Google Cloud.
+#  - NUXT_PUBLIC_GOOGLE_MAPS_API_KEY: client-visible, powers the map embed iframe.
+#    Ships to the browser by design — MUST be HTTP-referrer-restricted and scoped
+#    to the Maps Embed API in Google Cloud.
+NUXT_GOOGLE_MAPS_API_KEY=
 NUXT_PUBLIC_GOOGLE_MAPS_API_KEY=
--- a/README.md
+++ b/README.md
@@ -61,7 +61,8 @@ The variables the app actually reads:
 | `TRUSTED_ORIGINS` | optional | Comma-separated extra origins trusted for CSRF/Origin validation. |
 | `EMAIL_USER` | optional | Gmail address used to send notifications. |
 | `EMAIL_PASS` | optional | Gmail **app password** (not your account password — Gmail requires an app-specific password for SMTP). |
-| `NUXT_PUBLIC_GOOGLE_MAPS_API_KEY` | optional | Google Maps API key for the ride map / distance estimate features. Overrides `runtimeConfig.public.googleMapsApiKey`. |
+| `NUXT_GOOGLE_MAPS_API_KEY` | optional | **Server-only** Google Maps key for the Directions API distance/duration estimate. Never shipped to the browser. Overrides `runtimeConfig.googleMapsApiKey`. Restrict by IP/API in Google Cloud. |
+| `NUXT_PUBLIC_GOOGLE_MAPS_API_KEY` | optional | **Client-visible** Google Maps key for the ride map *embed* iframe. Ships to the browser by design, so it must be a **different** Cloud key that is **HTTP-referrer-restricted** and scoped to the Maps Embed API (issue #30). Overrides `runtimeConfig.public.googleMapsApiKey`. |
 
 Email is optional for local development: if `EMAIL_USER` / `EMAIL_PASS` are
 unset, sends fail and the failure is swallowed, so the app still runs and you
--- a/app/pages/rides/[id].vue
+++ b/app/pages/rides/[id].vue
@@ -193,6 +193,10 @@
     if (!ride.value) return ''
     const origin = encodeURIComponent(ride.value.pickupDisplay)
     const destination = encodeURIComponent(ride.value.dropoffDisplay)
+    // Issue #30: the PUBLIC key is correct here — the Maps Embed iframe runs in
+    // the browser and needs a client-visible key. It MUST be a DIFFERENT Google
+    // Cloud key from the server-only Directions key, HTTP-referrer-restricted
+    // and scoped to the Maps Embed API (owner action in Google Cloud Console).
     const apiKey = useRuntimeConfig().public.googleMapsApiKey
     return `https://www.google.com/maps/embed/v1/directions?key=${apiKey || ''}&origin=${origin}&destination=${destination}`
   })
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -4,7 +4,17 @@ export default defineNuxtConfig({
   modules: ['@nuxt/ui', 'nuxt-cron'],
   css: ['./assets/css/main.css'],
   runtimeConfig: {
+    // Server-only (issue #30): the Directions API key used by the estimate
+    // endpoint. It NEVER reaches the client bundle. Overridden by the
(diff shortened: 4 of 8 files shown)


TusharW4ni:
## 🚩 FLAGGED for human review — `nuxt.config.ts` risk file (fix verified correct & safe)

App-side of #30. Full `/issue-loop` verification (independent reviewer + a config/test-hygiene `/review` pass), then one round of test-hardening.

**Gate results:**
- **Suite (reviewer-run):** green — 35 files / 130 tests; `pnpm build` succeeds.
- **Revert-check:** PASS. Restoring the endpoint to read `config.public.googleMapsApiKey` makes `maps-key-split.test.ts` fail (`expected 'REQUEST_DENIED' to be 'Maps API Key not configured'`); the fix (reading the server-only key) passes.
- **Config split (verified correct):** `nuxt.config.ts` adds a server-only `googleMapsApiKey` (`NUXT_GOOGLE_MAPS_API_KEY`) alongside the public embed key; only `public.*` is serialized to the client bundle. The estimate endpoint reads `useRuntimeConfig(event).googleMapsApiKey`; grep confirms **no `server/` file reads the public Maps key** anymore. The client iframe still uses the public embed key (correct/required).
- **Scope/risk:** 8 files, tiny non-test footprint (~2 functional lines + config field; rest is docs/tests). Only risk file: `nuxt.config.ts`. No deps/CI/docker/schema/auth.

**Test-hygiene fix applied after first review:** the initial version set a bogus *server* key in the harness, which made cache-miss estimates fire real outbound calls to `maps.googleapis.com` — a flakiness/hang risk on no-egress CI. Inverted it: the harness now sets only the *public* key and leaves the server key empty, so post-fix the endpoint short-circuits at the "not-configured" gate and the green suite makes **zero outbound calls** (restored the estimate-cache exact assertion too).

**⚠️ Owner actions still required (out of scope here):**
1. **Rotate** the currently-exposed key (assume compromised — public repo + public bundle).
2. Provision **two different** Google Cloud keys: a private, API/IP-restricted **Directions** key → `NUXT_GOOGLE_MAPS_API_KEY` (server), and an HTTP-referrer-restricted, Embed-API-scoped key → `NUXT_PUBLIC_GOOGLE_MAPS_API_KEY` (client).
3. **Deployment note:** a deployment that currently sets only `NUXT_PUBLIC_GOOGLE_MAPS_API_KEY` will get `'Maps API Key not configured'` from the estimate feature until `NUXT_GOOGLE_MAPS_API_KEY` is also set — a breaking config change to apply at deploy time.

_Automated `/issue-loop` verification. Flagged per the risk-file rule; the app-side split is verified correct and hermetic. This is the in-repo half of #30 — the key rotation/restriction is yours._

PR #85, opened 2026-07-14
Fix #25: surface SMTP send failures instead of swallowing them

### LINKED ISSUE #25: [MEDIUM] Swallowed SMTP Errors Lead to False-Positive UI Feedback

TusharW4ni:
## What was broken

`sendEmail` swallowed every SMTP failure (only `console.error`) and returned nothing, so user-facing actions reported success even when the email never sent. In particular, `POST /api/auth/email-otp/send-verification-otp` returned 200 even when the OTP email failed — a false positive that left the user unable to sign in.

## What changed

- **`server/utils/email.ts`** — `sendEmail` now returns `Promise<boolean>`: `true` when sent, `false` when the send threw (the error is still logged).
- **`server/utils/auth.ts` (risk: auth OTP path)** — `sendVerificationOTP` now `await`s `sendEmail` and throws an `APIError('INTERNAL_SERVER_ERROR', ...)` when the send fails, so the OTP-send endpoint reports the real failure.
- **Background ride notifications stay fire-and-forget** (`server/utils/notification.ts`, `signup`/`unsignup`/`put rides`/`volunteers` welcome email) — they ignore the boolean and never throw into the request path, per #32. No functional change to those call sites.
- **Test-only bypass (strict prod no-op):** under `TEST_HOOKS=1` (only the e2e harness sets it) `sendEmail` is a strict no-op that reports success *without* sending, so `loginAs()` and every normal flow keep working with no SMTP. A test forces a failure by arming the `'email-send'` fault (reused `armFault`/`disarmFault` + new non-throwing `isFaultArmed` in `testHooks.ts`) via the new gated `POST /api/_test/arm-fault` endpoint (404 in prod, behind admin auth). When `TEST_HOOKS` is unset, none of this runs — production always really sends and surfaces real failures.

### Note on the admin "send test email" path
The owner decision mentions wiring an admin "send test email" endpoint. Grepping `sendEmail` across the codebase shows **no such endpoint exists** (the admin `notifications.vue` page only edits/toggles templates; there is no test-email button or route). Nothing to wire; not created here to avoid scope creep.

## Verification

- **Regression test:** `tests/e2e/otp-send-failure.test.ts`
  - `reports a failure when the OTP email cannot be sent` — arms the `email-send` fault, POSTs `send-verification-otp` for a seeded user, asserts a non-2xx response.
  - `a normal OTP send (no fault) still succeeds` — asserts 2xx (this is exactly what `loginAs` relies on).
- **Pre-fix FAIL (revert-check):** with `email.ts`/`auth.ts` reverted, the failure test fails with:
  `AssertionError: expected 200 to be greater than or equal to 400` (the swallowed send reports a false success). Post-fix it passes.
- **Full suite:** `pnpm test` → **35 files, 131 tests passed** (includes the new test; `loginAs` and the whole auth-gated suite still pass).
- **Build:** `pnpm build` succeeds.

Fixes #25

### DIFF
--- a/server/api/_test/arm-fault.post.ts
+++ b/server/api/_test/arm-fault.post.ts
@@ -1,0 +1,31 @@
+import { armFault, disarmFault } from '../../utils/testHooks'
+
+// Test-only endpoint (issues #25, #45). Arms or disarms a context-free fault
+// label so an e2e test can force a failure inside code that has no request
+// event of its own — e.g. sendEmail() in better-auth's sendVerificationOTP hook
+// (issue #25), which the OTP-send route surfaces as an error.
+//
+// Hard-gated: returns 404 unless TEST_HOOKS=1 (only the harness sets it), so it
+// does not exist in production. Still behind the global auth middleware (admin).
+//
+// Body: { label: string, disarm?: boolean }. The fault stays armed on the
+// server until disarmed, so a test must disarm it after use. Because the e2e
+// suite runs test files serially (vitest fileParallelism: false), no other
+// request overlaps the armed window.
+export default defineEventHandler(async (event) => {
+  if (process.env.TEST_HOOKS !== '1') {
+    throw createError({ statusCode: 404, statusMessage: 'Not found' })
+  }
+
+  const body = await readBody(event).catch(() => ({}))
+  const label = typeof body?.label === 'string' ? body.label : undefined
+  if (!label) {
+    throw createError({ statusCode: 400, statusMessage: 'label required' })
+  }
+
+  const disarm = body?.disarm === true
+  if (disarm) disarmFault(label)
+  else armFault(label)
+
+  return { ok: true, label, armed: !disarm }
+})
--- a/server/utils/auth.ts
+++ b/server/utils/auth.ts
@@ -59,7 +59,16 @@ export const auth = betterAuth({
   plugins: [
     emailOTP({
       async sendVerificationOTP({ email, otp, type }) {
-        sendEmail(email, 'OTP', `${otp}`)
+        // Surface send failures instead of swallowing them (issue #25): if the
+        // OTP email never sent, the user can't complete sign-in, so report the
+        // failure rather than a false success. `sendEmail` returns false on a
+        // failed send (and logs the error).
+        const sent = await sendEmail(email, 'OTP', `${otp}`)
+        if (!sent) {
+          throw new APIError('INTERNAL_SERVER_ERROR', {
+            message: 'Failed to send verification code. Please try again later.',
+          })
+        }
       },
     }),
   ],
--- a/server/utils/email.ts
+++ b/server/utils/email.ts
@@ -1,4 +1,5 @@
 import nodemailer from 'nodemailer'
+import { isFaultArmed } from './testHooks'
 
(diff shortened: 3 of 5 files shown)


TusharW4ni:
## 🚩 FLAGGED for human review — auth OTP path (fix verified correct, prod-safe, hardened)

Full `/issue-loop` verification (independent reviewer + a prod-safety/OTP `/review` pass), then one round of test-robustness hardening. **Flag is mandatory** because it touches `server/utils/auth.ts` (the OTP path) — not because of any defect.

**Gate results:**
- **Suite (reviewer-run):** green — 35 files / 131 tests; `pnpm build` succeeds; `/code-review` clean.
- **Revert-check:** PASS. Restoring `email.ts`+`auth.ts` to `origin/main` makes `otp-send-failure.test.ts` fail (`expected 200 to be greater than or equal to 400` — the swallowed send returning a false success); the fix passes.
- **Behavior:** `sendEmail` now returns a boolean; the OTP-send hook `await`s it and throws so `POST /send-verification-otp` reports a real failure instead of a false success. Ride notifications stay fire-and-forget (verified: `notification.ts`, signup/unsignup, `put/rides`, volunteer welcome all ignore the boolean and can't throw into the request path). There is no admin "send test email" endpoint in the codebase, so nothing else to wire.
- **Prod-safety of the test bypass (airtight — both reviewers):** `sendEmail`'s success no-op is strictly behind `TEST_HOOKS === '1'` (prod always really sends); `/api/_test/arm-fault` 404s unless `TEST_HOOKS=1` (and is admin-gated); `TEST_HOOKS` is set only in `tests/global-setup.ts` — absent from the dockerfile/CI/nuxt.config. Verified against better-auth internals that the OTP row is stored **before** the hook throws, so a retry works and `loginAs` still reads the code.
- **Scope:** 5 files, ~73 net non-test lines. Risk file: `server/utils/auth.ts`. No schema/CI/docker/deps/nuxt.config.

**Hardening applied after first review:** the fault disarm originally routed through `loginAs` (→ OTP-send), which was safe only via the session cache — a latent suite-wedge if a future `resetDb`/`clearSessionCache` landed mid-test. Reworked the test to capture the admin cookie once in `beforeAll` and arm/disarm via direct `appFetch` to the gated endpoint, so neither can ever be blocked by the armed fault. Test-only change; suite stayed green.

_Automated `/issue-loop` verification. Flagged per the risk-file rule; the fix is correct, the test bypass is prod-safe, and the fault seam is now wedge-proof._

PR #86, opened 2026-07-14
Fix #13: real server-side pagination on the list endpoints

### LINKED ISSUE #13: [HIGH] Unbounded Database Queries Without Pagination

TusharW4ni:
## What was broken

The four list endpoints (`server/api/get/{rides,clients,volunteers,admins}/index.ts`) ran unbounded `prisma.*.findMany()` calls — no `take`/`skip`. The rides dashboard fetched the **entire ride history** (with nested `client.user`/`volunteer.user` joins) on every load and filtered it client-side, so both the server and the browser degrade as the data grows (issue #13).

## What changed

**Backend — real server-side pagination (bounded).**
- New shared helper `server/utils/pagination.ts`: parses `?page` (1-based) and `?pageSize` (default 20), **hard-caps** `pageSize` to 100 and `page` to 1,000,000, and defines the `PageEnvelope` type.
- All four list endpoints now return `{ items, total, page, pageSize }` using `skip`/`take` + `count()` in a single `$transaction` (consistent total). Existing **soft-delete (`deletedAt: null`)** and **role-scoping** (issue #3) filters are preserved and applied to **both** the page query and the count.
- The rides dashboard's **date-range, status include/exclude, `assign:ME`, and text search** filtering moved to the backend (`?startDate`/`?endDate`/`?include`/`?exclude`/`?search`) so it composes with pagination instead of only filtering the current page. Status values are whitelisted against the real `RideStatus` enum so a bad param can't crash Prisma.

**Backend — bounded dropdown options endpoints.**
- `GET /api/get/clients/options` → `{ id, name, homeAddress }` (homeAddress kept for the create-ride pickup autofill), `GET /api/get/volunteers/options` → `{ id, name }`, AVAILABLE only. Both admin-only, hard-capped at 500 (bounded, not unbounded). The create/edit-ride volunteer pickers and the client picker now use these.

**Frontend — real pagination controls.**
- `app/pages/rides/index.vue` (rides table) and `app/pages/people.vue` (volunteers/clients/admins tabs) consume `.items`/`.total`, render `<UPagination>`, and **refetch server-side on page change**. Search is debounced; changing any filter/search/sort resets to page 1.

## Verification

**New regression test:** `tests/e2e/pagination.test.ts` (admin `loginAs`). Asserts, against the seeded 9 rides:
- `?page=1&pageSize=2` → `items.length === 2`, `total >= 9`, `page === 1`, `pageSize === 2`.
- `?page=2&pageSize=2` → a **different, non-overlapping** slice.
- oversized `?pageSize=100000` is **clamped to 100**.
- all four endpoints return the envelope; server-side `?include=status:CREATED` filtering yields `total === 4`.
- query-count seam stays bounded (`findMany` + `count`, not per-row).
- options endpoints return the minimal `{ id, name }` shape (3 AVAILABLE volunteers), no `user`/PII.
- empty `?pageSize=` falls back to the default (20); an absurd `?page=9007199254740991` is clamped (no unsafe-integer 500).

**Pre-fix failure (revert-checked):** with `server/api/get/rides/index.ts` reverted to the `origin/main` bare-array version, 6 of the pagination tests fail, e.g.:
> `expect(Array.isArray(res.items)).toBe(true)` — `Expected true, Received false`
> `TypeError: Cannot read properties of undefined (reading 'length')` at `expect(res.items.length)...`

because the old endpoint returns a bare array so `.items`/`.total` are `undefined`. Post-fix all pass.

**Contract change — every consumer updated.** This flips the four endpoints from a bare array to the envelope, so all consumers were migrated:
- Frontend: `people.vue`, `rides/index.vue`, `rides/[id].vue`.
- Tests (old array shape → `.items`, `pageSize=100` where the full set is needed): `smoke`, `query-count-seam` (bound raised 5→6 for the added `count()`), `pii-scoping`, `record-scoping`, `auth-middleware`, `signup-race`, `soft-deletes`, `safe-delete`, `orphaned-users`, `unassign-status`, `complete-requires-ridetime`, `audit-log`, `ride-cancel`, `ride-input-validation`, `ride-navigate-link`, `aaa-/zzz-order-dependence`, `topRiders-n1`, `address-normalization`, `address-matchkey-display`, `ride-broadcast-nonblocking`, `metrics-date-range`, `empty-phone`, `admin-empty-phone`, `admin-delete`.

**Suite:** `pnpm test` → **139 passed (35 files)**. `pnpm build` → success.

## Notes / decisions

- Frontend `.vue` behavior (UPagination, reactive refetch) can't be exercised by the e2e harness (API/SSR only); the backend pagination + options endpoints are fully tested, and the table UI is correct-by-inspection.
- **Search/filter moved server-side** for both pages (client-side filtering would only filter the current page once paginated). SQLite `contains` is case-insensitive for ASCII (matches the old lower-cased search); non-ASCII case-folding is a minor edge-case difference.
- The volunteer **options** endpoint is AVAILABLE-only. The edit-ride modal falls back to the ride's own volunteer record so an assigned-but-now-UNAVAILABLE volunteer still displays and isn't dropped on save.
- No schema/CI/docker/deps/auth-config changes.

Fixes #13

### DIFF
--- a/app/pages/people.vue
+++ b/app/pages/people.vue
@@ -10,17 +10,63 @@
 
   const { data: session } = await authClient.useSession(useFetch)
 
-  // --- Data Fetching ---
-  const { data: volunteers, refresh: refreshVolunteers } = await useFetch('/api/get/volunteers')
-  const { data: clients, refresh: refreshClients } = await useFetch('/api/get/clients')
-  const { data: admins, refresh: refreshAdmins } = await useFetch('/api/get/admins')
-
   // --- State ---
   const activeTab = ref('volunteers')
   const search = ref('')
   const volunteerStatusFilter = ref('All')
 
   const volunteerStatusOptions = ['All', 'AVAILABLE', 'UNAVAILABLE']
+
+  // --- Server-side pagination + search (issue #13) ---
+  // The roster endpoints are paginated, and search / status filtering now run on
+  // the server so they compose with pagination (client-side filtering would only
+  // filter the current page). Search is debounced and shared across the tabs.
+  const PAGE_SIZE = 20
+  const volunteerPage = ref(1)
+  const clientPage = ref(1)
+  const adminPage = ref(1)
+
+  const debouncedSearch = ref('')
+  const applyDebouncedSearch = debounce((value: string) => {
+    debouncedSearch.value = value
+  }, 300)
+  watch(search, (value) => applyDebouncedSearch(value))
+
+  const searchParam = computed(() => debouncedSearch.value || undefined)
+
+  // --- Data Fetching (reactive query -> auto refetch on page/search change) ---
+  const { data: volunteersData, refresh: refreshVolunteers } = await useFetch('/api/get/volunteers', {
+    query: {
+      page: volunteerPage,
+      pageSize: PAGE_SIZE,
+      search: searchParam,
+      status: computed(() =>
+        volunteerStatusFilter.value !== 'All' ? volunteerStatusFilter.value : undefined
+      ),
+    },
+  })
+  const { data: clientsData, refresh: refreshClients } = await useFetch('/api/get/clients', {
+    query: { page: clientPage, pageSize: PAGE_SIZE, search: searchParam },
+  })
+  const { data: adminsData, refresh: refreshAdmins } = await useFetch('/api/get/admins', {
+    query: { page: adminPage, pageSize: PAGE_SIZE, search: searchParam },
+  })
+
+  const volunteers = computed(() => volunteersData.value?.items ?? [])
+  const clients = computed(() => clientsData.value?.items ?? [])
+  const admins = computed(() => adminsData.value?.items ?? [])
+  const volunteerTotal = computed(() => volunteersData.value?.total ?? 0)
+  const clientTotal = computed(() => clientsData.value?.total ?? 0)
+  const adminTotal = computed(() => adminsData.value?.total ?? 0)
+
+  // Any query change returns the affected tab to page 1.
+  watch([debouncedSearch, volunteerStatusFilter], () => {
+    volunteerPage.value = 1
+  })
+  watch(debouncedSearch, () => {
+    clientPage.value = 1
+    adminPage.value = 1
+  })
 
   // --- Schemas ---
   const volunteerSchema = z.object({
@@ -99,43 +145,8 @@
     },
   })
 
-  // --- Computed ---
-  const filteredVolunteers = computed(() => {
-    if (!volunteers.value) return []
-
-    let result = volunteers.value
-
-    // Status Filter
-    if (volunteerStatusFilter.value !== 'All') {
-      result = result.filter((v: any) => v.status === volunteerStatusFilter.value)
-    }
-
-    // Search Filter
-    if (search.value) {
-      const q = search.value.toLowerCase()
-      result = result.filter(
-        (v: any) => v.user.name.toLowerCase().includes(q) || (v.user.email?.toLowerCase().includes(q) ?? false)
-      )
-    }
-
-    return result
-  })
-
-  const filteredClients = computed(() => {
-    if (!clients.value) return []
-    const q = search.value.toLowerCase()
-    return clients.value.filter(
-      (c: any) => c.user.name.toLowerCase().includes(q) || (c.user.email?.toLowerCase().includes(q) ?? false)
-    )
-  })
-
-  const filteredAdmins = computed(() => {
-    if (!admins.value) return []
-    const q = search.value.toLowerCase()
-    return admins.value.filter(
-      (a: any) => a.name.toLowerCase().includes(q) || (a.email?.toLowerCase().includes(q) ?? false)
-    )
-  })
+  // Search / status filtering is applied server-side (issue #13) so it composes
+  // with pagination; the tables bind directly to the returned page.
 
   const items = [
     {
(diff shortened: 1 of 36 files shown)


TusharW4ni:
### Review round 2 — addressed (commit 0ffa9dc)

**1. `VALID_STATUSES` dropped CANCELLED (confirmed regression) — fixed.** Added `'CANCELLED'` to the whitelist in `server/api/get/rides/index.ts` and removed the stale "CANCELLED is dead" comment. New regression test in `tests/e2e/pagination.test.ts` (`CANCELLED status filtering (#13 x #5)`): creates + cancels a throwaway ride, asserts `?exclude=status:COMPLETED,status:CANCELLED` drops it and `?include=status:CANCELLED` surfaces it, then soft-deletes it. **Pre-fix failure** (whitelist without CANCELLED): `expect(excluded.items.some((r) => r.id === rideId)).toBe(false)` → *"expected true to be false"* — the cancelled ride leaks past the exclude filter. Passes post-fix.

**2. Stable secondary sort — fixed.** Appended a unique `{ id: 'asc' }` tiebreaker to all four paginated `orderBy`s (rides: `[{ scheduledTime }, { id: 'asc' }]`; clients/volunteers/admins: `[{ <name> }, { id: 'asc' }]`) plus the two `options` endpoints, so tied `scheduledTime`/`name` rows can't be dropped or duplicated across pages.

**3. Volunteer assignment policy — restored.** `server/api/get/volunteers/options.ts` now returns ALL non-deleted volunteers (was AVAILABLE-only), matching main's behavior where the pickers used `/api/get/volunteers` (all statuses). #13 no longer narrows assignment policy. The edit-modal fallback in `[id].vue` stays but is no longer load-bearing.

**Suite:** `pnpm test` → **140 passed (35 files)**. `pnpm build` → success.

### Low-priority items (documented, not changed)
- **`options` endpoints hard-cap at 500 with no search.** Fine for the current roster sizes; if a deployment ever exceeds 500 clients/volunteers the pickers would truncate and should switch to a searchable/paginated selector then.
- **Server search uses Prisma `contains` → SQL `LIKE`.** `%`/`_` in the search term act as wildcards, and case-folding is ASCII-only — a minor drift from the old client-side lowercase `includes`. Not escaped here because Prisma's `contains` doesn't expose a `LIKE ... ESCAPE` clause for SQLite, so a partial pre-escape would be misleading; low severity (admin-only search box). Worth a dedicated follow-up if exact-substring matching is needed.

TusharW4ni:
## 🚩 FLAGGED for human review — large change (>200 lines, response-contract + frontend). Verified correct after 2 review rounds.

Full `/issue-loop` verification: independent reviewer (fresh worktree, suite + revert-check + empirical probing) + two `/review` finders (filtering-semantics, test-integrity/consumer-migration). One round of fixes applied and re-verified.

**Gate results (final):**
- **Suite (reviewer-run):** green — 35 files / **140 tests**; `pnpm build` succeeds.
- **Revert-check:** PASS. Restoring `rides/index.ts` to `origin/main` makes `pagination.test.ts` fail (`Array.isArray(res.items)` false + `Cannot read properties of undefined`); the new CANCELLED test also fails pre-fix (`expected true to be false` — cancelled ride leaks the exclude filter). Both pass with the fix.
- **Consumer migration — complete:** all table consumers read `.items`/`.total` (`rides/index.vue`, `people.vue` ×3 tabs); all dropdowns switched to the new bounded `/options` endpoints (`rides/index.vue`, `rides/[id].vue`). No consumer still expects a bare array.
- **Filtering semantics preserved:** role-scoping (volunteers see only own+CREATED, fail-closed), `assign:ME`, date range, search, and soft-delete (`deletedAt:null`) — with the SAME `where` applied to both `findMany` and `count()` in one `$transaction` (no total drift / no hidden-row count leak).
- **Test integrity:** all ~25 updated test files are legitimate envelope adaptations; no security/scoping assertion (pii-scoping, record-scoping, safe-delete, soft-deletes, admin-delete) was weakened. `query-count-seam` 5→6 justified by the added `count()`.

**Three issues found and fixed this review (all re-verified):**
1. **CANCELLED filter regression (was breaking #5):** a stale `VALID_STATUSES` whitelist dropped `status:CANCELLED`, so cancelled rides leaked into the default dashboard and the Cancelled filter was a no-op. Empirically caught (no test covered it — seed has no cancelled rides). Fixed + added a `CANCELLED status filtering (#13 × #5)` regression test.
2. **Pagination correctness — no stable sort:** added a unique `{ id: 'asc' }` tiebreaker to all four paginated `orderBy`s (and the options endpoints), so `skip`/`take` can't skip or duplicate rows on tied `scheduledTime`/`name` values.
3. **Unintended assignment-policy change:** `volunteers/options` was AVAILABLE-only, blocking assigning an unavailable volunteer to a new ride; restored to all non-deleted volunteers (main's behavior).

**Design:** envelope `{ items, total, page, pageSize }` (default 20, cap 100), `skip`/`take`+`count()` in a `$transaction`; rides+people search/status/date filtering moved server-side; dedicated bounded `/options` endpoints (`requireAdmin`, cap 500, minimal shape) for the dropdowns; `<UPagination>` on the rides + people tables.

**Notes for the reviewer:** net non-test +295 (>200); the frontend `<UPagination>` wiring is correct-by-inspection (the API/SSR harness can't drive client reactivity) — the backend pagination + options endpoints are fully test-covered. Documented low-priority follow-ups: the `/options` 500-cap has no search (typeahead would be the scale fix), and server search uses `LIKE` so `%`/`_` act as wildcards + ASCII-only case-folding (Prisma doesn't expose `LIKE ... ESCAPE` for SQLite).

_Automated `/issue-loop` verification (find → fix → re-verify). Flagged for size + frontend-reactivity; the implementation is verified correct._

PR #102, opened 2026-07-14
Fix #88: guard DELETE /api/delete/admins/[id] (role check, self-delete, last-admin)

### LINKED ISSUE #88: DELETE /api/delete/admins/[id] archives ANY user (no role check), and allows self-delete / deleting the last admin
## Symptoms (all verified live)
1. **No role check**: `DELETE /api/delete/admins/<userId>` looks up `user` by id with no `role: 'ADMIN'` filter (`server/api/delete/admins/[id].ts`). Pointing it at a VOLUNTEER's user id returns 200 and archives the user row **while their `volunteer` row stays active** — an inconsistent half-archived state that also bypasses the invariants of `delete/volunteers` (ASSIGNED rides are NOT released back to the pool, unlike a proper volunteer delete).
2. **Self-delete allowed**: an admin can archive their own account via their own session (verified: disposable admin self-deleted → 200).
3. **No last-admin guard**: nothing prevents archiving the final ADMIN. Since only admins can create admins (`POST /api/post/admins` is admin-only), this can permanently lock the org out of user management.

## Suggested fix
In `server/api/delete/admins/[id].ts`:
- Require the target user to have `role: 'ADMIN'` (404 otherwise).
- Reject self-delete (400/409) — or at minimum reject when `count(active admins) == 1`.

## Repro
```
# as admin, take any volunteer's user.id
DELETE /api/delete/admins/<volunteer-user-id>   → 200, user archived, volunteer profile + assigned rides untouched
```

TusharW4ni:
## What was broken

`DELETE /api/delete/admins/[id]` looked up the target `user` by id with **no `role: 'ADMIN'` filter**, so pointing it at a non-admin (e.g. a VOLUNTEER's underlying `user.id`) returned **200 and archived that user** while the `volunteer` row stayed active and their ASSIGNED rides were never released — an inconsistent half-archived state that bypasses the invariants of `delete/volunteers`. The endpoint also allowed an admin to **self-delete** and to **delete the last remaining admin**, which can permanently lock the org out of user management (only an admin can create/promote admins via the admin-only `POST /api/post/admins`).

## What changed (`server/api/delete/admins/[id].ts` only)

- **Role check (headline):** scope the `findFirst` lookup to `role: 'ADMIN'` (and mirror it on the `tx.user.update` `where` for defense-in-depth), so a non-admin id resolves to "not found" → **404**. This stops archiving volunteers/clients through the admin endpoint.
- **Reject self-delete:** resolve the caller via `getAuth(event)`; if `id === session.user.id`, throw **409** before the transaction.
- **Reject last-admin:** a fast pre-transaction check (`count(active admins) <= 1` → 409) for the common case, **plus** a re-count *inside* the write-serialized transaction after archiving that rolls back with the same 409 if the delete would leave zero admins.

Everything the endpoint already did correctly is preserved: soft-delete via archive (deletedAt + release email/phone to deletedEmail/deletedPhone then null them), `session.deleteMany` to revoke sessions, the `writeAuditLog('ADMIN_DELETED', …)` call, and the P2025 → 404 handling. A legit delete of a different, non-last admin still returns **200** and archives.

## Verification

New regression test: **`tests/e2e/delete-admins-guard.test.ts`** (4 tests):

1. `refuses to archive a non-admin (volunteer) user → 404, volunteer untouched` — **headline; fails pre-fix, passes post-fix.**
2. `rejects an admin deleting their own account → 409, account stays active` — self-delete guard, isolated from the last-admin guard (a disposable 2nd admin + 2 seeded admins are present). Fails pre-fix.
3. `still archives a different, non-last admin → 200 and it is gone` — non-regression; passes pre- **and** post-fix, proving the guards don't over-block.
4. `cannot delete the last remaining admin (lockout prevented) → 409` — reduces to a single admin, then that admin's self-delete is rejected.

**Pre-fix run** (fix reverted, test present): 3 failed / 1 passed. Exact headline failure:

```
FAIL tests/e2e/delete-admins-guard.test.ts > refuses to archive a non-admin (volunteer) user → 404, volunteer untouched
AssertionError: deleting a non-admin via the admin endpoint must 404: expected 200 to be 404 // Object.is equality
- 404
+ 200
```

(Tests 2 and 4 also failed pre-fix with `expected 200 to be 409`; test 3 passed pre-fix as designed.)

**Post-fix run:** whole suite green — **38 files / 147 tests passed** (baseline 37/143 + 4 new).

### Notes / decisions

- **Self-delete policy:** always rejected (409), independent of how many admins exist — an admin cannot delete their own account through this endpoint.
- **Last-admin cannot be isolated from self-delete at the API level:** only an admin can call this endpoint, so deleting "the last admin" is necessarily the caller deleting themselves; the self-delete guard fires first and emits the 409. The count-based guard is therefore belt-and-suspenders. Test 4 proves the end-to-end lockout guarantee but does not isolate the count guard (documented in the test).
- **Concurrency:** the pre-transaction count alone has a TOCTOU (two admins deleting *each other* concurrently are not self-deletes and could both read count > 1). The in-transaction re-count after archiving closes it on the write-serialized better-sqlite3 adapter. This path isn't unit-tested because the harness serializes writes and the realistic lockout is already prevented by the self-delete guard.

### Risk files touched

None. The change is confined to `server/api/delete/admins/[id].ts` plus the new test file — no middleware, schema, CI, Docker, or dependency changes.

## Follow-ups

None. Sibling issues noted only where relevant in comments; no scope expansion.

Fixes #88

### DIFF
--- a/server/api/delete/admins/[id].ts
+++ b/server/api/delete/admins/[id].ts
@@ -11,23 +11,61 @@ export default defineEventHandler(async (event) => {
     })
   }
 
+  // Issue #88: reject self-delete. An admin archiving their own account (and, in
+  // particular, the last admin — see the count guard below) can permanently lock
+  // the org out of user management, since only an ADMIN can create/promote
+  // admins (POST /api/post/admins is admin-only). The caller's session is
+  // resolved by the global auth middleware onto event.context.auth.
+  const session = getAuth(event)
+  if (session?.user?.id === id) {
+    throw createError({
+      statusCode: 409,
+      statusMessage: 'You cannot delete your own admin account',
+    })
+  }
+
+  // Issue #88: this endpoint deletes ADMINs only. Without the role filter it
+  // would archive ANY user (e.g. a VOLUNTEER's user row), leaving a half-archived
+  // inconsistent state — the volunteer profile stays active and their ASSIGNED
+  // rides are never released, unlike a proper delete/volunteers. Scope the lookup
+  // to role: 'ADMIN' so a non-admin id resolves to "not found" (404).
+  //
   // Soft delete (issue #27): archive the user instead of destroying it. Read the
   // current contact values so we can RELEASE them: copy email -> deletedEmail and
   // phone -> deletedPhone, then null out email/phone so the same email/phone can
   // be reused by a new record without a unique violation (decision #2). Guard on
   // deletedAt: null so a second delete 404s (P2025) rather than re-archiving.
   const existing = await prisma.user.findFirst({
-    where: { id, deletedAt: null },
+    where: { id, role: 'ADMIN', deletedAt: null },
     select: { email: true, phone: true },
   })
   if (!existing) {
     throw createError({ statusCode: 404, statusMessage: 'Admin not found' })
   }
 
+  // Issue #88: never delete the last remaining admin. In practice this is almost
+  // always a self-delete (only an admin can call this, and the sole admin
+  // deleting themselves is caught above), so this early check is mainly a clear,
+  // fast error for the common sequential case. It is NOT sufficient on its own:
+  // two admins deleting EACH OTHER concurrently are not self-deletes and could
+  // both read count > 1 here before either archive commits — so the invariant is
+  // re-checked inside the write-serialized transaction below.
+  const activeAdmins = await prisma.user.count({
+    where: { role: 'ADMIN', deletedAt: null },
+  })
+  if (activeAdmins <= 1) {
+    throw createError({
+      statusCode: 409,
+      statusMessage: 'Cannot delete the last remaining admin',
+    })
+  }
+
   try {
     const updated = await prisma.$transaction(async (tx) => {
       const u = await tx.user.update({
-        where: { id, deletedAt: null },
+        // role: 'ADMIN' mirrors the findFirst guard above (issue #88,
+        // defense-in-depth) so this path can never archive a non-admin.
+        where: { id, role: 'ADMIN', deletedAt: null },
         data: {
           deletedAt: new Date(),
           deletedEmail: existing.email,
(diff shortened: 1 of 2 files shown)


TusharW4ni:

Independently verified by a fresh reviewer subagent (separate worktree, no author context) + an orchestrator `/review` pass. Auto-merge was attempted but **blocked by the environment's production-deploy safety guard** (agents cannot merge to `main`) — *not* by any quality gate. **Ready for a maintainer to squash-merge.**

| Gate | Result |
|---|---|
| Reviewer-run full suite | ✅ **38 files / 147 tests pass**, 0 fail (foreground `pnpm test`) |
| Revert-check (test pins the bug) | ✅ **Non-vacuous** — with the fix reverted, 3/4 new tests fail (non-admin delete `200≠404`; self-delete `200≠409`; last-admin `200≠409`). Restored → **4/4 pass**. |
| Scope | ✅ 2 files: `server/api/delete/admins/[id].ts` + new test. **Net non-test: +53 lines.** |
| Risk files | ✅ **None** (no schema/middleware/auth-config/CI/Docker/deps). |
| Findings (reviewer + `/review`) | ✅ **Zero blocking.** |
| Priority | P2 security (not P0) → auto-merge eligible. |

### What the fix does & why it's correct
Adds three guards to `DELETE /api/delete/admins/[id]`, confined to that handler:
- **Role scope** (`role:'ADMIN'` on both the lookup and the tx update) → a volunteer/client id now 404s instead of half-archiving the user (leaving their profile + assigned rides dangling). Verified: the targeted volunteer stays `VOLUNTEER` and in-roster.
- **Self-delete → 409**, before any write (unbypassable — only an admin session reaches the handler).
- **Last-admin → 409** pre-check **plus an in-transaction re-count backstop** for the concurrent cross-delete race.

**The concurrency backstop was verified at the adapter level, not hand-waved:** `@prisma/adapter-better-sqlite3@7.2.0` serializes interactive transactions via a single mutex on one connection, so two admins deleting each other resolve to exactly one archive surviving — no interleaving reaches zero admins. The 409 thrown inside `$transaction` rolls back both the archive and the session revoke; it's re-thrown as a genuine 409 (not masked as P2025→404). `writeAuditLog('ADMIN_DELETED')` fires only on a committed transaction.

**Non-blocking suggestion (optional):** add a `Promise.all` cross-delete test to pin the TOCTOU invariant directly (currently proven by reasoning + adapter inspection). No code change required to merge.

Issue #89, opened 2026-07-14
Partial PUT on people endpoints silently wipes email/phone (clients, volunteers, admins; post/clients too)


TusharW4ni:
## Symptom
Any people-update request that omits a contact field **nulls it** instead of leaving it unchanged. Verified live:

- `PUT /api/put/clients/[id]` with body `{"name":"X"}` → the client's **email AND phone are set to NULL**.
- `PUT /api/put/volunteers/[id]` with `{"name":"X"}` → **phone set to NULL** (email survives because it's passed through as `undefined`).
- `PUT /api/put/admins/[id]` with `{"name":"X"}` → **phone set to NULL**.
- `POST /api/post/clients` with an **existing** user's email (upsert branch) also runs `phone: emptyToNull(body.phone)` → wipes that user's stored phone.

## Root cause
Two patterns in `server/api/put/clients/[id].ts`, `put/volunteers/[id].ts`, `put/admins/[id].ts`, `post/clients/index.ts`:
- `email: body.email || null` — missing field coerces to `null` (explicit wipe) instead of `undefined` (no-op).
- `phone: emptyToNull(body.phone)` — `emptyToNull(undefined)` returns `null` (see `server/utils/sanitize.ts`), so an omitted phone is an explicit wipe.

## Impact
The current people-page edit modal happens to send a full payload (and normalizes phone before save — verified in browser), so the UI doesn't trigger it today. But any partial API call (script, future UI change, mobile client) silently destroys contact data. Defense-in-depth bug.

## Suggested fix
Only include a field in the Prisma `data` when it is present in the body (`if ('phone' in body) data.phone = emptyToNull(body.phone)`), or use a zod schema with `.optional()` + explicit-null semantics (pairs well with the validation gap issue).

Found during the 2026-07-14 exhaustive audit.

PR #106, opened 2026-07-14
Fix #92: reject cross-type profile creates (enforce singular roles)

### LINKED ISSUE #92: post/clients & post/volunteers with an existing email create dual profiles / silently change roles
## Symptoms (verified live)
- `POST /api/post/clients` with an existing **VOLUNTEER's** email → 200. The user now has BOTH a `volunteer` and a `client` profile while `role` stays `VOLUNTEER`. The client half is invisible to the role model, and `delete/volunteers` (which assumes \"roles are singular\" — see its comment) archives the shared `user` row, stranding the client profile.
- `POST /api/post/volunteers` with an existing **CLIENT's** email → 200, and the user's role is silently upgraded CLIENT → VOLUNTEER while their client profile and rides remain. A client gains volunteer API access without any explicit admin decision surfaced in the UI.

## Suggested fix
Decide the intended model. If roles are singular (everything else assumes it): reject creates when the email belongs to a user with a different active profile (409 with a clear message). If dual roles are intended, the role model, deletes, and scoping all need to handle it — much bigger change.

## Repro
```
POST /api/post/clients    {"name":"Bob","email":"bob@example.com", ...address}  → 200, volunteer+client dual profile (also wipes bob's phone — see #89)
POST /api/post/volunteers {"name":"George","email":"george@example.com"}        → 200, role CLIENT→VOLUNTEER
```

TusharW4ni:
## What was broken

Roles are singular everywhere in the app, but the two people-create endpoints let a single user acquire two profile types:

- `POST /api/post/clients` with an **existing volunteer's** email returned **200** and left the user with BOTH a volunteer and a client profile while `role` stayed `VOLUNTEER`. The client half was invisible to the role model, and `delete/volunteers` would strand it (the `#89` phone-wipe also fired on the upsert branch).
- `POST /api/post/volunteers` with an **existing client's** email returned **200** and **silently upgraded** `role` CLIENT→VOLUNTEER (`role: user.role === 'CLIENT' ? 'VOLUNTEER' : user.role`) while the client profile and its rides remained.

## Decision — singular roles

The issue notes "everything else assumes roles are singular." This PR enforces that on create: a create is **rejected with 409** when the email belongs to a user who already has a *different* active profile type. `post/admins` is intentionally unchanged (promoting an existing user to ADMIN is intended — R-080).

## What changed

- `server/api/post/clients/index.ts`: include the `volunteer` relation on the lookup; in the existing-user branch, **409** "That email belongs to a volunteer/admin account" when the user has an active (`deletedAt: null`) volunteer profile **or** role `VOLUNTEER`/`ADMIN`. An existing CLIENT (or a plain user with no profile) is unaffected and still falls through to the existing "already a client" **400** (R-042) / normal create.
- `server/api/post/volunteers/index.ts`: include the `client` relation; **409** "That email belongs to a client account" when the user has an active client profile **or** role `CLIENT`, instead of upgrading. Dropped the silent CLIENT→VOLUNTEER role bump. An existing VOLUNTEER still hits "already a volunteer" **400** (R-061).

Both guards throw before any DB mutation (and before the address upsert on the client path), so a rejected create leaves the target account untouched. The UI (`app/pages/people.vue`) already toasts `statusMessage`, so the 409 message surfaces cleanly.

## Pins flipped

- `tests/e2e/known-bugs.test.ts`: R-048 pin flipped `it.fails(` → `it(`, and added an adjacent R-048 case for the **post/volunteers** direction (existing CLIENT `george@example.com` → 409).
- `REQUIREMENTS.md`: R-048 status `pin [#92]` → `auto`.

## Verification

- `pnpm test` → **43 files, 189 passed | 3 expected-fail | 0 failed** (baseline was 187/4/0; +1 from the flipped R-048, +1 from the new post/volunteers case).
- `pnpm trace` → **119/119 covered ✔** (R-048 now counts as auto: 109 auto / 4 pin).
- `pnpm build` → succeeds.
- **Revert-check**: with the two server files reverted, both flipped R-048 tests fail as expected — pre-fix assertion:
  `× R-048 (#92): creating a client with an existing VOLUNTEER email is rejected` and `× R-048 (#92): creating a volunteer with an existing CLIENT email is rejected` (both returned 200, expected `[400,409]`). They pass with the fix.

## Follow-up (out of scope)

`POST /api/post/volunteers` with an **admin's** email still creates a volunteer profile for the admin (role kept ADMIN) — a pre-existing admin dual-profile edge the issue scopes to the CLIENT direction; left for a separate decision rather than widening this diff.

Fixes #92

### DIFF
--- a/REQUIREMENTS.md
+++ b/REQUIREMENTS.md
@@ -65,7 +65,7 @@ is not documentation that can rot — it is enforced.
 | R-045 | A partial update (`{name}` only) must NOT wipe email/phone — omitted fields are no-ops | `put/clients/[id].ts` et al | auto | `known-bugs.test.ts` |
 | R-046 | Update/delete of an unknown or archived client → 404 | `put/clients/[id].ts` | auto | `requirements-flows.test.ts` |
 | R-047 | Setting a client's email to another user's email → clean 4xx conflict, not 500 | `put/clients/[id].ts` | auto | `known-bugs.test.ts` |
-| R-048 | People creates on an email that belongs to a different active profile type are rejected (no dual profiles, no silent role upgrades) | `post/clients`, `post/volunteers` | pin [#92](https://github.com/UTDallasEPICS/wcoa/issues/92) | `known-bugs.test.ts` |
+| R-048 | People creates on an email that belongs to a different active profile type are rejected (no dual profiles, no silent role upgrades) | `post/clients`, `post/volunteers` | auto | `known-bugs.test.ts` |
 | R-049 | Deleting a client with active (non-deleted) rides is blocked with 409 (#23) | `delete/clients/[id].ts` | auto | `safe-delete.test.ts` |
 | R-050 | Client soft-delete archives user+client, releases email/phone to `deletedEmail/deletedPhone` (reusable), revokes sessions, hides from lists, preserves metrics history (#27) | `delete/clients/[id].ts` | auto | `soft-deletes.test.ts` |
 
--- a/server/api/post/clients/index.ts
+++ b/server/api/post/clients/index.ts
@@ -28,7 +28,8 @@ export default defineEventHandler(async (event) => {
   let user = null
   if (email) {
     user = await prisma.user.findUnique({
-      where: { email }
+      where: { email },
+      include: { volunteer: true }
     })
   }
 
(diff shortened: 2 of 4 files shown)


TusharW4ni:

Auto-merge attempted-and-blocked only by the production-deploy guard (agents can't merge `main`). **Ready for a maintainer to squash-merge.**

| Gate | Result |
|---|---|
| Full suite | ✅ **189 passed / 3 expected-fail / 0 failed** (independently confirmed by the reviewer subagent + the author) |
| Revert-check | ✅ **Non-vacuous** — reverting the two handlers makes **both** R-048 tests fail (client-with-volunteer-email and volunteer-with-client-email return 200 instead of 409); restored → 13 passed / 0 failed |
| Traceability (`pnpm trace`) | ✅ 119/119 covered |
| Scope | ✅ 2 handlers + `known-bugs.test.ts` + `REQUIREMENTS.md`, **~+29 net non-test** |
| Risk files | ✅ **None** |
| Findings | ✅ No blocking correctness/security issues |

### What it does
Enforces the singular-role model: `post/clients` 409s an email that belongs to a volunteer/admin; `post/volunteers` 409s an email that belongs to a client and **drops the silent CLIENT→VOLUNTEER role upgrade**. Legit paths preserved (new email creates; same-type re-post hits the existing "already a client/volunteer" 400; soft-deleted profiles don't false-block since a released email can't be matched).

### Verification note (transparency)
The independent reviewer subagent **stalled on an infra watchdog** *after* independently confirming the full suite green (189/0-failed). The orchestrator (not the author) then completed the revert-check (via a detached PR-head checkout) and the adversarial pass — so author↔verifier independence holds.

### Minor / deferred (NOT a blocker, out of #92's stated scope)
Asymmetry: `post/clients` rejects an ADMIN email, but `post/volunteers` does **not** — an admin's email → `post/volunteers` still creates a volunteer profile (admin+volunteer dual, role stays ADMIN). #92's symptoms list only the client↔volunteer cases (both fixed here). Closing the admin edge is a trivial one-condition follow-up (`|| user.role === 'ADMIN'` on the `post/volunteers` guard) if you want the model fully symmetric.

Issue #87, opened 2026-07-14
Volunteers cannot complete their own rides — UI offers 'Mark as Completed' but the API returns 403


TusharW4ni:
## Symptom
As a VOLUNTEER, open one of your ASSIGNED rides (`/rides/[id]`) → the page shows **Unsign Up** and **Mark as Completed**. Fill the completion modal (Total Duration) and submit → toast \"Failed to complete ride\"; the ride stays ASSIGNED.

## Root cause
The completion modal calls `PUT /api/put/rides/[id]` (app/pages/rides/[id].vue), but the global auth middleware (`server/middleware/auth.ts`) only allows volunteer writes matching `VOLUNTEER_WRITE_ALLOW` (signup, unsignup, `put/volunteers/bySession/**`). `PUT /api/put/rides/**` is not in the allowlist → **403 Admin access required**.

Verified live (prod build, volunteer session):
```
PUT /api/put/rides/da532fa0-… {"status":"COMPLETED","totalRideTime":1.5} → 403 Admin access required
```

## Impact
The volunteer ride-completion feature is entirely broken — only admins can complete rides. Completed-hours metrics depend on volunteers reporting `totalRideTime`, so this blocks the core workflow.

## Suggested fix (either)
1. Add a scoped allowlist entry + handler check: volunteers may `PUT` **their own assigned ride** with only `{status: 'COMPLETED', totalRideTime}` (mirror the unsignup ownership check); or
2. Add a dedicated `POST /api/post/rides/[id]/complete` self-service endpoint (consistent with signup/unsignup), and point the UI at it.

Option 2 keeps the middleware simple and the whitelist honest.

## Notes
- The e2e suite exercises ride completion with an ADMIN session only, which is why this regression of the volunteer flow is invisible to CI. A regression test should log in as a volunteer.
- Found during the 2026-07-14 exhaustive post-backlog audit (all backend routes probed as all four roles + full browser walkthrough).

PR #108, opened 2026-07-15
Fix #95: reject invalid ride status transitions out of terminal states

### LINKED ISSUE #95: No server-side ride status-transition guard (COMPLETED→CREATED un-completes a ride, keeps stale totalRideTime)
## Symptom (verified live)
`PUT /api/put/rides/[id]` accepts any status transition. e.g. a COMPLETED ride (totalRideTime=1.5) can be sent `{"status":"CREATED"}` → 200. The ride re-enters the available signup pool while keeping its stale `totalRideTime`; completion metrics silently change. CANCELLED→ASSIGNED etc. are also accepted.

This was explicitly flagged as an owner judgment call in PR #77 (\"no server-side status-transition guard — possible follow-up issue\") but never filed.

## Suggested fix
Validate transitions in `server/api/put/rides/[id].ts` against a small state machine, e.g. allow CREATED↔ASSIGNED, ASSIGNED→COMPLETED, CREATED/ASSIGNED→CANCELLED; reject un-completing/un-cancelling (or clear `totalRideTime`/notify when deliberately supported). Admin-only surface, hence P3.

TusharW4ni:
## What was broken
`PUT /api/put/rides/[id]` accepted **any** status transition. A COMPLETED ride (with a `totalRideTime` set) sent `{"status":"CREATED"}` returned `200`: it silently re-entered the volunteer signup pool while keeping its stale `totalRideTime`, corrupting completion metrics. `CANCELLED→ASSIGNED` and other un-cancels were likewise accepted. Flagged as an owner judgment call in PR #77 but never filed until #95.

## The fix (minimal, safe terminal-state guard)
Added a guard immediately after the `existingRide` null-check in `server/api/put/rides/[id].ts`. `COMPLETED` and `CANCELLED` are terminal: an **explicit** status change to a **different** status is rejected with **409 Conflict** (consistent with the codebase's other state-conflict responses — #87/#88/#92).

```ts
const TERMINAL = ['COMPLETED', 'CANCELLED']
if (body.status !== undefined && body.status !== existingRide.status && TERMINAL.includes(existingRide.status)) {
  throw createError({ statusCode: 409, statusMessage: `Cannot change the status of a ${existingRide.status.toLowerCase()} ride` })
}
```

Why this is minimal and low-risk:
- Fires **only** when `body.status` is sent **and** differs from the current status, so **same-status no-op edits** (e.g. editing a COMPLETED ride's notes/totalRideTime with `status:'COMPLETED'`) stay allowed.
- The implicit `ASSIGNED→CREATED` unassign auto-reset (#7) runs only when `body.status` is `undefined`, so the guard never touches it.
- Forward transitions `CREATED↔ASSIGNED`, `ASSIGNED→COMPLETED`, `CREATED/ASSIGNED→CANCELLED` all still work.
- The #10 completion guard, #14 cache invalidation, #28 audit logging, and notifications are untouched.

Net non-test change: **+11 lines of code** (+9 comment) in one route handler. The fuller state-machine suggested in the issue is deliberately deferred — this directly fixes the reported bug with the least regression surface.

## Pin flipped
- `tests/e2e/known-bugs.test.ts`: R-108 `it.fails(` → `it(`, assertion widened `expect(res.status).toBe(400)` → `expect([400, 409]).toContain(res.status)` (409 chosen).
- `REQUIREMENTS.md`: R-108 `pin` → `auto`.

## Necessary test-hygiene adaptation (please note)
Four ride test files recycled seed rides through the **exact** `COMPLETED→CREATED` transition this fix now (correctly) rejects, in their `afterEach` teardown or fixture selection. Because the R-108 pin requires that transition to be *rejected*, these teardowns can no longer perform it — there is no guard formulation that both rejects the bug and lets the teardown do it (a `volunteerId`-based exemption would just be an exploitable loophole). The teardowns are made **tolerant** (non-throwing `appFetch`, mirroring the existing `ride-cancel.test.ts` pattern), and `ride-input-validation` now selects a non-terminal ride for its valid-update case. **No feature assertions changed** — only cleanup/fixture mechanics:
- `complete-requires-ridetime.test.ts` (R-107), `unassign-status.test.ts` (R-106), `rides-volunteer-complete.test.ts` (#87): `afterEach` reset via non-throwing `appFetch`.
- `ride-input-validation.test.ts` (R-105): `firstRideId` picks a `CREATED`/`ASSIGNED` ride.

## Verification
- `pnpm test`: **43 files, 190 passed / 2 expected-fail / 0 failed** (baseline on `origin/main` was 189 / 3 / 0; R-108 flipped from expected-fail to passing).
- Explicitly confirmed still green: R-105 (`ride-input-validation`), R-106 (`unassign-status`), R-107 (`complete-requires-ridetime`), R-109 (`ride-cancel`), #87 (`rides-volunteer-complete`), and the requirements-flows ride lifecycle.
- `pnpm trace`: **119/119 covered** (R-108 now `auto`; 3 pins remain).
- `pnpm build`: succeeds.
- **Revert-check:** with the handler guard removed but the flipped test in place, R-108 fails with `AssertionError: expected [ 400, 409 ] to include 200` (buggy handler returns 200); with the guard restored it passes. Fail-before / pass-after confirmed.

Fixes #95

### DIFF
--- a/REQUIREMENTS.md
+++ b/REQUIREMENTS.md
@@ -105,7 +105,7 @@ is not documentation that can rot — it is enforced.
 | R-105 | Ride update accepts ONLY whitelisted fields; unknown keys 400 via `.strict()` (mass-assignment guard, #31) | `put/rides/[id].ts` | auto | `ride-input-validation.test.ts` |
 | R-106 | Unassigning a volunteer (volunteerId → null/"") auto-resets status to CREATED — but only when the ride was ASSIGNED and no explicit status was sent (#7) | `put/rides/[id].ts` | auto | `unassign-status.test.ts` |
 | R-107 | Completing a ride requires totalRideTime ≥ 0.1 (on the request or already stored) (#10) | `put/rides/[id].ts` | auto | `complete-requires-ridetime.test.ts` |
-| R-108 | Ride status transitions are guarded server-side (no COMPLETED→CREATED un-complete with stale totalRideTime) | `put/rides/[id].ts` | pin [#95](https://github.com/UTDallasEPICS/wcoa/issues/95) | `known-bugs.test.ts` |
+| R-108 | Ride status transitions are guarded server-side (no COMPLETED→CREATED un-complete with stale totalRideTime) | `put/rides/[id].ts` | auto | `known-bugs.test.ts` |
 | R-109 | Cancelling notifies the previously-assigned volunteer via the RIDE_CANCELLED template (#5) | `put/rides/[id].ts` | auto | `ride-cancel.test.ts` |
 | R-110 | Completing notifies the volunteer (RIDE_COMPLETED) and emails admins | `put/rides/[id].ts` | auto | `requirements-flows.test.ts` |
 | R-111 | Update/complete/signup against an archived ride → 404 (#27) | `put/rides`, signup | auto | `soft-deletes.test.ts` |
--- a/server/api/put/rides/[id].ts
+++ b/server/api/put/rides/[id].ts
@@ -56,6 +56,28 @@ export default defineEventHandler(async (event) => {
     })
   }
 
+  // Issue #95: COMPLETED and CANCELLED are terminal states. Reject an explicit
+  // status change *out* of one of them. e.g. sending {status:'CREATED'} to a
+  // COMPLETED ride would re-enter it into the signup pool while keeping its
+  // stale totalRideTime, silently corrupting completion metrics; CANCELLED→
+  // ASSIGNED is likewise wrong. This fires only when the caller explicitly
+  // sends a *different* status, so same-status no-op edits (e.g. editing a
+  // COMPLETED ride's notes with status:'COMPLETED') stay allowed, and the
+  // implicit ASSIGNED→CREATED unassign auto-reset below — which runs only when
+  // body.status is undefined — is untouched. Forward transitions (CREATED↔
+  // ASSIGNED, ASSIGNED→COMPLETED, CREATED/ASSIGNED→CANCELLED) all still work.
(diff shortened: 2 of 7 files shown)


TusharW4ni:

Blocked from auto-merge only by the production-deploy guard (agents can't merge `main`). **Ready for a maintainer to squash-merge.**

| Gate | Result |
|---|---|
| Full suite (reviewer-run) | ✅ **43 files, 190 passed / 2 expected-fail / 0 failed** |
| Revert-check | ✅ Reverting **only the handler** makes **exactly R-108 fail** (`expected [400,409] to include 200`), everything else green → the guard is load-bearing **and** the test edits mask no regression |
| Scope | ✅ Non-test source = only `put/rides/[id].ts` (**~+11 code**); the 4 test-file edits are necessary neighbors |
| Risk files | ✅ **None** |
| Findings | ✅ None |

### About the test-file edits (the sensitive part — cleared)
The PR touches 4 existing test files because their `afterEach` cleanup / fixtures recycled seed rides via the exact `COMPLETED→CREATED` transition the fix now correctly rejects. Both reviewers audited every change: **3 are `afterEach` throwing-`$fetch`→non-throwing-`appFetch` (identical body), 1 is fixture selection (pick a non-terminal ride), 1 is the R-108 pin flip. No `it()` assertion was removed or weakened.** The revert-check independently proves they hide nothing: those files pass against BOTH the old and new handler, so they're insensitive to whether the reset succeeds.

### What it does
Terminal-state guard in `put/rides/[id]`: an explicit status change *out of* COMPLETED/CANCELLED → **409**. Same-status edits, the implicit unassign auto-reset (#7/R-106), and all forward transitions are untouched. Closes the un-complete/un-cancel metric-corruption hole.

Issue #91, opened 2026-07-14
People update endpoints turn routine conflicts into 500s (unknown id on put/admins, duplicate email on any people update)


TusharW4ni:
## Symptoms (verified live)
- `PUT /api/put/admins/<unknown-id>` → **500** (unhandled Prisma P2025). Siblings (`put/clients`, `put/volunteers`, `put/rides`, all deletes) correctly 404 — `put/admins/[id].ts` is the only update endpoint with no existence check.
- `PUT /api/put/clients/[id]` setting `email` to another user's email → **500** (unhandled P2002 unique violation). Same for the other people updates. Should be a 409/400 with a clear \"email already in use\" message.

## Suggested fix
- `put/admins`: `findFirst({id, deletedAt: null})` → 404, like its siblings.
- Wrap people-update user writes with a P2002 catch → 409 \"email/phone already in use\".

Found during the 2026-07-14 exhaustive audit.

TusharW4ni:
One more endpoint for this class, found while building the regression framework: `PUT /api/put/notifications/templates/<unknown-name>` with a **valid** body → 500 (unhandled P2025 from `prisma.notificationTemplate.update`). With an invalid body it 400s on validation first, which is why the audit's hand probe missed it. Pinned in `tests/e2e/known-bugs.test.ts` (R-190 pin).

PR #110, opened 2026-07-15
Fix #105: scope put/admins/[id] to ADMIN targets (404 for non-admins)

### LINKED ISSUE #105: PUT /api/put/admins/[id] has no role:'ADMIN' filter — can edit any user by id (sibling of #88)
## Symptom
`PUT /api/put/admins/[id]` looks up its target with `findFirst({ id, deletedAt: null })` — **no `role: 'ADMIN'` filter**. So an admin can point the "admins" update endpoint at a **CLIENT or VOLUNTEER's user id** and change that user's name/email/phone through it. This is the update-side analogue of the delete-side bug fixed in #88 (`delete/admins` now requires `role: 'ADMIN'`).

## Context
- Discovered during the #91 fix (PR #104). That PR **added** the existence check (unknown id → 404) that this endpoint previously lacked — so PR #104 *tightens* behavior (the base did a bare `prisma.user.update({ where: { id } })` with no check at all). It deliberately did **not** add the role filter, to stay in scope for #91 (error-mapping), hence this follow-up.
- Impact is low today: the endpoint is admin-only (via the global middleware), and admins can already edit people through the other endpoints. It's an authorization-consistency / defense-in-depth gap, not a privilege escalation. (Also note the broader no-auth caveat of #1.)

## Suggested fix
Scope the lookup (and the update `where`) to `role: 'ADMIN'`, mirroring `delete/admins/[id].ts` from #88 — a non-admin target → 404. Add a regression pin (`R-###`) in `known-bugs.test.ts` / `REQUIREMENTS.md`.

Found while fixing #91 (2026-07-14).

TusharW4ni:
## What was broken

`PUT /api/put/admins/[id]` looked up its target with `findFirst({ id, deletedAt: null })` — **no `role: 'ADMIN'` filter** — and then updated `where: { id }`. So an admin could edit a CLIENT's or VOLUNTEER's user row by id, bypassing `put/clients` / `put/volunteers` and their own validation. This is the update-side analogue of #88 (which scoped `delete/admins` to ADMIN).

## What changed

`server/api/put/admins/[id].ts` (+11/-2):
- Added `role: 'ADMIN'` to the existence-check `findFirst` → a non-admin id now resolves to 404 "Admin not found".
- Added `role: 'ADMIN', deletedAt: null` to the `prisma.user.update` `where` too (defense-in-depth, mirroring `delete/admins/[id].ts`). A concurrent role-change/soft-delete makes the write match no row → P2025 → 404 via the existing `throwOnPrismaWriteConflict`.
- Unchanged: the #91 P2002→409 / P2025→404 mapping and the #89/#90 presence-aware field building.

## Regression pin

- `REQUIREMENTS.md`: new row **R-085** — "`put/admins` only edits users whose role is ADMIN (404 otherwise), mirroring delete/admins (#105)", source `put/admins/[id].ts`, status `auto`, coverage `known-bugs.test.ts`.
- `tests/e2e/known-bugs.test.ts`: `R-085 (#105): put/admins refuses to edit a non-admin user` — as admin, resolve bob's (VOLUNTEER) userId via `/api/get/volunteers/bySession`, snapshot his name, `PUT /api/put/admins/<userId>` `{name:'…'}` → asserts **404** and that the volunteer's user name is unchanged.

## Verification

- `pnpm test` → **192 passed | 1 expected-fail | 0 failed** (baseline + my new passing test).
- `pnpm trace` → **covered 120/120 ✔** (R-085 added).
- `pnpm build` → succeeds.
- **Revert-check**: with `put/admins/[id].ts` reverted to `origin/main`, the new test fails with `AssertionError: expected 200 to be 404` (the endpoint returns 200 and renames the volunteer). With the fix it passes and the user is untouched. Fail-before / pass-after confirmed.
- `/code-review` (xhigh) → no findings.

Fixes #105

### DIFF
--- a/REQUIREMENTS.md
+++ b/REQUIREMENTS.md
@@ -92,6 +92,7 @@ is not documentation that can rot — it is enforced.
 | R-082 | `delete/admins` only archives users whose role is ADMIN (404 otherwise) | `delete/admins/[id].ts` | auto | `known-bugs.test.ts` |
 | R-083 | An admin cannot delete their own account, and the last remaining admin cannot be deleted (lockout guard) | `delete/admins/[id].ts` | auto | `known-bugs.test.ts` |
 | R-084 | Admin soft-delete releases email/phone, revokes sessions; double-delete 404s | `delete/admins/[id].ts` | auto | `admin-delete.test.ts`, `soft-deletes.test.ts` |
+| R-085 | `put/admins` only edits users whose role is ADMIN (404 otherwise), mirroring delete/admins (#105) | `put/admins/[id].ts` | auto | `known-bugs.test.ts` |
 
 ## 6. Rides — CRUD & lifecycle
 
--- a/server/api/put/admins/[id].ts
+++ b/server/api/put/admins/[id].ts
@@ -30,8 +30,14 @@ export default defineEventHandler(async (event) => {
   // prisma.user.update and surfaced the raw P2025 as a 500. 404 like the sibling
   // update endpoints (put/clients, put/volunteers). Soft delete (issue #27): an
   // archived user is treated as gone.
+  //
+  // Issue #105: this endpoint edits ADMINs only — the update-side analogue of
+  // #88 on delete/admins. Without the role filter it would edit ANY user (e.g. a
+  // CLIENT's or VOLUNTEER's user row) by id, bypassing put/clients / put/volunteers
+  // and their own validation. Scope the lookup to role: 'ADMIN' so a non-admin id
+  // resolves to "not found" (404), mirroring delete/admins/[id].ts.
   const existing = await prisma.user.findFirst({
-    where: { id, deletedAt: null },
+    where: { id, role: 'ADMIN', deletedAt: null },
   })
   if (!existing) {
     throw createError({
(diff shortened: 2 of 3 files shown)


TusharW4ni:

Blocked from auto-merge only by the production-deploy guard. **Ready to squash-merge.**

| Gate | Result |
|---|---|
| Full suite (orchestrator-run) | ✅ **43 files, 192 passed / 1 expected-fail / 0 failed** |
| Revert-check | ✅ Reverting `put/admins/[id].ts` makes the new **R-085** pin fail (`expected 200 to be 404` — endpoint edited the volunteer); restored → passes |
| Traceability / build | ✅ 120/120 · build OK |
| Scope | ✅ 1 handler + new R-085 pin + REQUIREMENTS row (~+11 net non-test) |
| Risk files | ✅ **None** |
| Findings | ✅ None — a textbook mirror of the merged #88 `delete/admins` fix (`role:'ADMIN'` on the `findFirst` and `update` where) |

Closes the update-side authorization-consistency gap: `PUT /api/put/admins/[id]` now 404s for a non-admin target instead of editing any user by id.

PR #112, opened 2026-07-15
Fix #98: resolve SSR hydration mismatch in date rendering

### LINKED ISSUE #98: Vue hydration mismatch (2×) on every page load
## Symptom
Every SSR page load (dashboard, /rides, /admin/notifications — reproduced on all) logs two console errors:
```
[error] Hydration completed but contains mismatches.
```
Consistent 2-per-load across different pages points at a layout-level component (nav/theme toggle or a date rendered with locale-dependent formatting) rather than page content. Prod build, Chrome, 2026-07-14 audit.

## Impact
No user-visible breakage observed, but hydration mismatches force client re-renders and can mask real reactivity bugs; they also make the console noisy for anyone debugging.

## Suggested fix
Reproduce with `nuxt build && node .output/server/index.mjs`, open any page with devtools; Vue prints the mismatching component in dev mode (`nuxt dev`) — track down whichever node differs between server and client render (usual suspects: `new Date()`/locale formatting during SSR, `prefers-color-scheme`-dependent markup).

TusharW4ni:
## What was broken

Every SSR page that renders a date logged `[error] Hydration completed but contains mismatches.` This is a **frontend/SSR** fix.

**Root cause (exact node):** the pages formatted dates with bare `new Date(x).toLocaleString()` / `toLocaleString('en-US', { … })` and **no fixed `timeZone`**. `toLocaleString` reads the timezone (and, with no locale arg, the locale) from the host, so the server (UTC in the prod container) and the viewer's browser (their local TZ) turn the *same instant* into *different strings*. Concretely the `/rides` **"Date" column `<td>`** rendered `Jul 19, 2026, 04:52 AM` on a UTC server vs `Jul 18, 2026, 11:52 PM` in a US‑Central browser — Vue sees the text differ and re-renders on the client. Same pattern in `rides/[id].vue` (appointment + pickup time) and `admin/audit.vue` (createdAt).

**Why it looked layout-level / "2× / intermittent":** it only manifests when the server TZ ≠ browser TZ (prod UTC vs local browser), so a same‑TZ build shows nothing (matching the prior R‑304 note). Vue caps the mismatch log to one message per app instance, and that single message **persists in the console across the app's client-side (SPA) route changes**, so after visiting `/rides` it appeared to linger on pages that render no dates (dashboard, notifications). The "2×" is the devtools double-reporting a single log (observed directly).

## The fix

New client helper `app/utils/datetime.ts` — `formatDateTime(value, options?)` pins the locale (`en-US`) and timezone (`America/Chicago`, the org's zone, matching the server-side convention in `server/utils/datetime.ts` from #9). Server and client now format identically → clean hydration. Each call site keeps its own date/time style. Non-test change: **+35 / -6 across 4 files** (`app/utils/datetime.ts` new; `rides/index.vue`, `rides/[id].vue`, `admin/audit.vue` swapped to the helper).

## Verification

**Empirical browser before/after (prod build: `node .output/server/index.mjs`, server `TZ=UTC`, browser `America/Chicago`):**
- **BEFORE** (origin/main code, fresh `/rides` load): console showed
  ```
  [error] Hydration completed but contains mismatches.
  [error] Hydration completed but contains mismatches.
  ```
  and the "Date" cell was `Jul 19, 2026, 04:52 AM` (SSR/UTC) vs `Jul 18, 2026, 11:52 PM` (client/Central).
- **AFTER** (this branch, identical conditions): console **clean** (verified capture was live), and SSR + client both render `Jul 18, 2026, 11:52 PM`.

**Regression test:** `tests/e2e/hydration-datetime.test.ts` — a pure-function pin (cf. #9's `notification-datetime.test.ts`) asserting the helper renders an instant in `America/Chicago` **regardless of host TZ** (verified identical output under `TZ=UTC` and `TZ=Asia/Tokyo`). It fails to import pre-fix (helper absent), so a revert of this commit turns it red.

**Suite + build:** `pnpm test` → 44 files, 194 passed + 1 pre-existing expected-fail. `pnpm build` succeeds.

**Browser pin (R‑304):** tightened from a non-gating diagnostic (`expect(true).toBe(true)`) to a hard assertion (`expect(hydration).toEqual([])`) in `tests/browser/admin-flows.spec.ts`; `REQUIREMENTS.md`/`TESTING.md` updated. **UNVERIFIED here** — Playwright (`pnpm test:browser`) isn't runnable in this environment. Note the Playwright suite runs the Nuxt server and browser on the same machine (same TZ), so it does not itself exercise the UTC-vs-local divergence; the server-side determinism is what `hydration-datetime.test.ts` guards.

**Evidence level:** empirical browser before/after achieved (via a deliberately induced server-TZ≠browser-TZ) + inspection argument + green suite. The one gap is the live Playwright R‑304 run.

Fixes #98

### DIFF
--- a/REQUIREMENTS.md
+++ b/REQUIREMENTS.md
@@ -208,7 +208,7 @@ is not documentation that can rot — it is enforced.
 | R-301 | Settings (volunteer): profile card, status selector, per-type notification toggles, reminder editor — all persisted via bySession endpoints | `app/pages/settings.vue` | auto | browser: `volunteer-flows.spec.ts` |
 | R-302 | Admin notification page: 5 template cards with Edit/Disable reflecting `enabled` | `app/pages/admin/notifications.vue` | auto | browser: `admin-flows.spec.ts` |
 | R-303 | Admin audit page: bounded table, action/user filters | `app/pages/admin/audit.vue` | auto | browser: `admin-flows.spec.ts` |
-| R-304 | Pages hydrate without mismatches (SSR output == client render) | layout | manual [#98](https://github.com/UTDallasEPICS/wcoa/issues/98) | browser: `admin-flows.spec.ts` non-gating diagnostic (mismatch reproduces intermittently; observed in audit) |
+| R-304 | Pages hydrate without mismatches (SSR output == client render): SSR-rendered dates are timezone-stable so server (UTC) and browser (local TZ) agree | `app/utils/datetime.ts`, rides/index.vue, rides/[id].vue, admin/audit.vue | auto [#98](https://github.com/UTDallasEPICS/wcoa/issues/98) | `hydration-datetime.test.ts` (server-side determinism) + browser: `admin-flows.spec.ts` (`/rides` hydrates clean) |
 | R-305 | Success/error toasts on every mutating UI action | all pages | auto | browser specs (asserted on key flows) |
 
 ## Known non-requirements / accepted behavior
--- a/TESTING.md
+++ b/TESTING.md
@@ -31,7 +31,9 @@ untagged requirement turns CI red** — that's what keeps the catalog honest ins
 The 2026-07-14 audit found 12 open defects (issues #87–#98). Each has a pin that asserts the
 **correct** behavior, wrapped so today's wrong behavior is "expected":
 - API pins use `it.fails(...)` in `known-bugs.test.ts`.
-- UI pins use `test.fail()` in the browser specs (R-135/#87, and the R-304/#98 diagnostic).
+- UI pins use `test.fail()` in the browser specs (R-135/#87). The R-304/#98 hydration check is now a
+  hard assertion (`/rides` hydrates clean) — its server-side determinism is guarded by
+  `tests/e2e/hydration-datetime.test.ts`.
 
 **When you fix a bug, its pin starts failing** (because the assertion now passes). That is the signal to:
 1. remove the `.fails` / `test.fail()`, and
--- a/app/pages/admin/audit.vue
+++ b/app/pages/admin/audit.vue
@@ -76,7 +76,7 @@ function formatDetails(details: unknown): string {
             :key="log.id"
             class="border-t border-gray-100 dark:border-gray-800"
           >
-            <td class="px-3 py-2 whitespace-nowrap">{{ new Date(log.createdAt).toLocaleString() }}</td>
+            <td class="px-3 py-2 whitespace-nowrap">{{ formatDateTime(log.createdAt) }}</td>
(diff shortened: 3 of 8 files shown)


TusharW4ni:
## 🚩 issue-loop gate: FLAGGED — frontend fix, well-evidenced but not automatically revert-checkable

**Why flagged (not a quality doubt):** this is a **frontend SSR-hydration** fix. The loop's automated gate runs the vitest API/SSR harness, which **cannot revert-check a hydration mismatch** — that lives in the Playwright browser layer (`pnpm test:browser`), which isn't runnable in this environment. The R-304 browser guard was tightened to a hard assertion but is **unverified here**. So the fix can't clear the automated revert-check gate, even though the evidence below is strong.

### Verification achieved
| Check | Result |
|---|---|
| Full suite (orchestrator-run) | ✅ **44 files, 194 passed / 1 expected-fail / 0 failed** |
| **Empirical browser before/after (author)** | ✅ Prod build, server `TZ=UTC` + browser `America/Chicago`: BEFORE (origin/main) `/rides` logged the hydration error ×2 (cell `Jul 19 04:52 AM` SSR vs `Jul 18 11:52 PM` client); AFTER: console clean, both render `Jul 18 11:52 PM` |
| New unit test | ✅ `hydration-datetime.test.ts` pins the helper's TZ-determinism (host-TZ-independent; checked under `TZ=UTC` and `TZ=Asia/Tokyo`) |
| Risk files | ✅ **None** (`nuxt.config.ts` NOT touched) |

### Root cause & fix
SSR date sites used bare `toLocaleString()` with no fixed timezone → server (UTC container) and client (viewer-local) formatted the same instant into different strings → hydration mismatch on every date-rendering page. Fix: a new `app/utils/datetime.ts#formatDateTime` pinning `en-US` + `America/Chicago`, used by `rides/index.vue`, `rides/[id].vue`, `admin/audit.vue`. **The timezone matches the existing `server/utils/datetime.ts` (#9)**, so client display and server-side emails now agree org-wide.

### Two things for your call
1. **Run `pnpm test:browser`** before/at merge to confirm the R-304 browser assertion passes (the loop can't run Playwright here).
2. **Behavior change:** displayed dates are now always in **America/Chicago** (the org's zone), not the viewer's local timezone. For a Texas org tracking rides in its own zone this is the consistent/correct choice and matches the notification emails — but non-Central viewers will see Central time. Confirm that's intended.

**Follow-up (separate, noted by author):** server-side email formatters (`signup.ts`/`unsignup.ts`/`scheduler.ts`) still use bare `toLocaleString()` — not a hydration issue (email bodies), but a latent server-TZ correctness gap worth its own ticket.

